PR #1, opened 2023-07-27
docs(storybook): use new addon-designs plugin

brian-patrick-3:
## Summary

Replaced `storybook-addon-designs` with the new official `@storybook/addon-designs` plugin for embedding Figma designs.


maulik-thaker1 (approved):
LGTM( Looks good to me ). 
There are some Lint warnings.
Just curious why we removed `import { withDesign } from 'storybook-addon-design';`  from each story files ?

PR #8, opened 2023-08-28
feat(breadcrumbs): 1st version

peterpham-kyndryl:
## Summary

A new Breadcrumbs component per design https://www.figma.com/file/6AovH7Iay9Y7BkpoL5975s/Component-Library-for-Dev?node-id=459%3A39107&mode=dev

https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/1238385

## Notes

- Detailed change notes
- can go here

## To Do

- Anything else
- left to do

## Checklist

## Testing Instructions

## Screenshots

(if any)

### DIFF
--- a/src/components/reusable/breadcrumbs/breadcrumbs.stories.ts
+++ b/src/components/reusable/breadcrumbs/breadcrumbs.stories.ts
@@ -1,0 +1,57 @@
+/**
+ * Copyright Kyndryl, Inc. 2023
+ */
+
+// External library imports
+import { html } from 'lit';
+import type { Meta, StoryObj } from '@storybook/web-components';
+
+import '@kyndryl-design-system/foundation/components/link';
+
+// Relative component imports
+import './breadcrumbs';
+import './index';
+
+/**
+ * ## Usage
+ *
+ * Here's how you can use the `kyn-breadcrumbs` component with sample items:
+ *
+ * ```javascript
+ *
+ * html`
+ *  <kyn-breadcrumbs>
+ *    <kd-link class="breadcrumb-item" .href="${'/'}">Bridge</kd-link>
+ *    <kd-link class="breadcrumb-item" .href="${'/catalog'}">Catalog</kd-link>
+ *    <span class="breadcrumb-item breadcrumb-last">FinOps</span>
+ *  </kyn-breadcrumbs>
+ * `;
+ * ```
+ *
+ * Remember to import and define the component before using it.
+ */
+
+const meta: Meta = {
+  title: 'Components/Breadcrumbs',
+  component: 'kyn-breadcrumbs',
+  parameters: {
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/file/6AovH7Iay9Y7BkpoL5975s/Component-Library-for-Dev?node-id=459%3A39033&mode=dev',
+    },
+  },
+};
+
+export default meta;
+
+type Story = StoryObj;
+
+export const Breadcrumbs: Story = {
+  render: () => {
+    return html`<kyn-breadcrumbs>
+      <kd-link class="breadcrumb-item" .href="${'/'}">Bridge</kd-link>
+      <kd-link class="breadcrumb-item" .href="${'/catalog'}">Catalog</kd-link>
+      <span class="breadcrumb-item breadcrumb-last">FinOps</span>
+    </kyn-breadcrumbs>`;
+  },
+};
--- a/src/components/reusable/breadcrumbs/breadcrumbs.ts
+++ b/src/components/reusable/breadcrumbs/breadcrumbs.ts
@@ -1,0 +1,36 @@
+/**
+ * Copyright Kyndryl, Inc. 2023
+ */
+
+// External library imports
+import { html, LitElement } from 'lit';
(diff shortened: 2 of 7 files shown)


src/components/reusable/breadcrumbs/breadcrumbs.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+ * ```javascript
+ *
+ * html`
+ *  <kyn-breadcrumbs>
+ *    <kd-link class="breadcrumb-item" .href="/">Bridge</kd-link>
+ *    <kd-link class="breadcrumb-item" .href="/catalog">Catalog</kd-link>
+ *    <span class="breadcrumb-item breadcrumb-last">FinOps</span>
+ *  </kyn-breadcrumbs>
+ * `;
+ * ```
```
brian-patrick-3: I would avoid using Lit templates in any examples that aren't auto-generated, since the consumer is most likely not using Lit.

src/components/reusable/breadcrumbs/breadcrumbs.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+export default meta;
+
+type Story = StoryObj;
+
+export const Breadcrumbs: Story = {
+  render: () => {
+    return html`<kyn-breadcrumbs>
+      <kd-link class="breadcrumb-item" .href="${'/'}">Bridge</kd-link>
+      <kd-link class="breadcrumb-item" .href="${'/catalog'}">Catalog</kd-link>
+      <span class="breadcrumb-item breadcrumb-last">FinOps</span>
```
brian-patrick-3, line 54: Should be able to style these elements without needing to add a class by targeting the tag name instead. I would not put the onus on app devs to add these specific classes.

src/components/reusable/breadcrumbs/breadcrumbs.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// Relative component imports
+import stylesheet from './breadcrumbs.scss';
+
+/**
+ * Breadcrumbs Component.
+ *
+ * @description Provides a navigation aid that displays the user's location in a hierarchical structure.
+ *
+ * @slot - Slot for inserting breadcrumb items, typically links indicating the navigation path.
```
brian-patrick-3, line 17: Need to name the slot in the comment for it to show in the props table. Use "unnamed" in this case.

src/components/reusable/breadcrumbs/breadcrumbs.ts
```diff
@@ -0,0 +1,36 @@
...
+import { html, LitElement } from 'lit';
+import { customElement } from 'lit/decorators.js';
+
+// Relative component imports
+import stylesheet from './breadcrumbs.scss';
+
+/**
+ * Breadcrumbs Component.
+ *
+ * @description Provides a navigation aid that displays the user's location in a hierarchical structure.
```
brian-patrick-3, line 15: I don't think `@description` is supported by the CEM analyzer. Just put the description where it says "Breadcrumb component."

brian-patrick-3 (changes requested):
Looking better with the `<slot>` approach, though still one thing that could be problematic with the slotted link class names. 

I did notice there is some extra spacing around the dashes caused by whitespace. Should try to remove that so it's even on each side (16px).

src/components/reusable/breadcrumbs/breadcrumbs.stories.ts
brian-patrick-3: You're right, they could technically put anything in the slot and there's not much we can do about it. We just have to provide the guidelines and examples, but asking them to add specific classes is unnecessary and confusing,

peterpham-kyndryl:
> Looking better with the `<slot>` approach, though still one thing that could be problematic with the slotted link class names.
> 
> I did notice there is some extra spacing around the dashes caused by whitespace. Should try to remove that so it's even on each side (16px).

the extra space is from the Link itself.

brian-patrick-3:
> the extra space is from the Link itself.
Hmm you're right, will have to fix that in Foundation then.

marcosebastianlopez (approved):

brian-patrick-3:
> > the extra space is from the Link itself.
> 
> Hmm you're right, will have to fix that in Foundation then.

@peterpham-kyndryl Try pulling in foundation beta.23 and see if it fixes the link whitespace.

src/components/reusable/breadcrumbs/breadcrumbs.stories.ts
peterpham-kyndryl: Hmm we're trying to give them much flexibility with the slot approach but then restrict them with certain tags. It feels strange/hard-coded, especially if later on we have our own components such as <kyn-link> or <Typography> that consuming apps want to put in the slot.

brian-patrick-3: That's just how subcomponents work with slots in a web component ecosystem.

peterpham-kyndryl:
Work is done in another PR https://github.com/kyndryl-design-system/web-components/pull/11

PR #10, opened 2023-08-29
fix: text inputs validation nesting

brian-patrick-3:
## Summary

Fix the if/else structure for validation on text input and text area components. Previously it was not validating min/max unless required = true.


maulik-thaker1 (approved):
LGTM

PR #16, opened 2023-10-02
fix(dropdown): remove height when closed

brian-patrick-3:
## Summary

Dropdown options were taking up space on the page when hidden due to CSS visibility. Set height to 0 to prevent this.


CunningDesigns (approved):
approved..

PR #19, opened 2023-10-09
fix(table): fix sorting

peterpham-kyndryl:
## Summary

Link here (if applicable).

## Notes

- Detailed change notes
- can go here

## To Do

- Anything else
- left to do

## Checklist

## Testing Instructions

## Screenshots

(if any)

### DIFF
--- a/src/components/reusable/table/table-head.ts
+++ b/src/components/reusable/table/table-head.ts
@@ -15,6 +15,23 @@ import styles from './table-head.scss';
 export class TableHead extends LitElement {
   static override styles = [styles];
 
+  override firstUpdated() {
+    this.addEventListener('on-sort-changed', this.handleChildSort as EventListener);
+  }
+
+  private handleChildSort(e: CustomEvent) {
+    const sortedColumnKey = e.detail.sortKey;
+
+    // Get all kyn-th elements
+    const allHeaders = Array.from(this.querySelectorAll('kyn-th'));
+
+    for (const header of allHeaders) {
+      if (header.sortKey !== sortedColumnKey) {
+        header.resetSort(); // Reset sort state of non-sorted columns
+      }
+    }
+  }
+
   override render() {
     return html` <slot></slot> `;
   }
--- a/src/components/reusable/table/table-head.ts
+++ b/src/components/reusable/table/table-head.ts
@@ -1,5 +1,5 @@
 import { html, LitElement } from 'lit';
-import { customElement } from 'lit/decorators.js';
+import { customElement, queryAssignedElements } from 'lit/decorators.js';
 
 import styles from './table-head.scss';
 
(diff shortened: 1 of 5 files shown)


src/components/reusable/table/table-head.ts
```diff
@@ -15,6 +15,23 @@ import styles from './table-head.scss';
 export class TableHead extends LitElement {
   static override styles = [styles];
 
+  override firstUpdated() {
+    this.addEventListener('on-sort-changed', this.handleChildSort as EventListener);
+  }
```
brian-patrick-3, line 20: Add event listeners in `connectedCallback()` instead. Also need to remove event listeners in `disconnectedCallback()`, otherwise they can start to duplicate themselves.

src/components/reusable/table/table-head.ts
```diff
@@ -15,6 +15,23 @@ import styles from './table-head.scss';
...
+
+  private handleChildSort(e: CustomEvent) {
+    const sortedColumnKey = e.detail.sortKey;
+
+    // Get all kyn-th elements
+    const allHeaders = Array.from(this.querySelectorAll('kyn-th'));
```
brian-patrick-3, line 26: I'm surprised this works, since the `kyn-th` are slotted. Normally you would want to use this Lit decorator: https://lit.dev/docs/api/decorators/#queryAssignedElements

src/components/reusable/table/table-header.ts
```diff
@@ -68,11 +96,34 @@ export class TableHeader extends LitElement {
...
+  override updated() {
+    this.checkIfHeaderSlotIsEmpty();
+  }
+
+  checkIfHeaderSlotIsEmpty() {
+    // Retrieve the slot from the shadow DOM.
+    const slot = this.shadowRoot!.querySelector('slot');
+
+    // Get all nodes assigned to the slot.
+    const nodes = slot!.assignedNodes({ flatten: true });
```
brian-patrick-3, line 115: It might be easier to use the Lit decorator for this: https://lit.dev/docs/api/decorators/#queryAssignedNodes

src/components/reusable/table/table-header.ts
```diff
@@ -86,10 +137,28 @@ export class TableHeader extends LitElement {
...
+    const ariaLabel =
+      this.sortable && this.headerLabel
+        ? `Sort by ${this.headerLabel}`
+        : undefined;
+
     return html`
       <div
         class="container"
         @click=${this.sortable ? () => this.toggleSortDirection() : undefined}
+        role=${role}
```
brian-patrick-3, line 151: You can use Lit's ifDefined decorator here: https://lit.dev/docs/api/directives/#ifDefined

brian-patrick-3 (changes requested):
I like the more realistic example much better. I still think sorting needs to cycle back to "default" state, but we can revisit that later. One comment about binding event listeners needs to be addressed, otherwise just added some suggestions to use Lit decorators and directives in a few places.

src/components/reusable/table/table-head.ts
peterpham-kyndryl: Right, I forgot the connectedCallback. Was still thinking in React.

src/components/reusable/table/table-head.ts
```diff
@@ -15,15 +15,33 @@ import styles from './table-head.scss';
...
   private handleChildSort(e: CustomEvent) {
     const sortedColumnKey = e.detail.sortKey;
 
-    // Get all kyn-th elements
-    const allHeaders = Array.from(this.querySelectorAll('kyn-th'));
+    // unnamedSlotEls[0] is the kyn-tr element
+    const parentRow = this.unnamedSlotEls[0];
+
+    // Get all kyn-th children of that kyn-tr element
+    const allHeaders = Array.from(parentRow.querySelectorAll('kyn-th'));
```
brian-patrick-3, line 44: See comment about using selector in the decorator above. You should be able to loop through `unnamedSlotEls` directly if done correctly.

brian-patrick-3 (changes requested):
A couple minor things related to the `@queryAssignedElements()` usage

src/components/reusable/table/table-head.ts
```diff
@@ -15,15 +15,33 @@ import styles from './table-head.scss';
 export class TableHead extends LitElement {
   static override styles = [styles];
 
-  override firstUpdated() {
-    this.addEventListener('on-sort-changed', this.handleChildSort as EventListener);
+  @queryAssignedElements()
+  unnamedSlotEls!: Array<HTMLElement>;
```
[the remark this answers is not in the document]
peterpham-kyndryl: By selector, you mean a css selector?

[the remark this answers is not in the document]
brian-patrick-3: Yeah, ex: `@queryAssignedElements({ selector: '.selector' })`. I just tried it though, and it didn't seem to pull in the nested slots unfortunately.  Even using the `flatten` option, despite what the MDN docs say:

> If the `flatten` option is set to `true`, it returns a sequence of both the elements assigned to this slot, as well as the elements assigned to any other slots that are descendants of this slot. If no assigned elements are found, it returns the slot's fallback content.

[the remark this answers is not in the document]
peterpham-kyndryl: I did try both using kyn-th as an selector and giving it a unique css class to act as a selector; with flatten is set to true but I'm getting an from assignedElements function.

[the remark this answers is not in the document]
brian-patrick-3: Yeah doesn't seem to work unfortunately, oh well!

brian-patrick-3 (approved):

PR #24, opened 2023-10-26
feat(table): implement phase 2 enhancements

peterpham-kyndryl:
## Summary

Link here (if applicable).

## Notes

- Detailed change notes
- can go here

## To Do

- Anything else
- left to do

## Checklist

## Testing Instructions

## Screenshots

(if any)


brian-patrick-3 (changes requested):
1. One thing I notice is the table-toolbar is included within the overflow-x table-container, which means the table title and buttons can scroll off screen. Only the table itself should be in the overflow container. This is noticeable on the Basic Table story, but some stories seem to have nested table-containers.
   1. The "Sticky Header" story suffers here too, there seem to be multiple vertical scrollbars.
3. Maybe combine the batch actions from the story called "Data Table" into the Multi Select story, since those two things are interconnected.
4. The inline actions should support a "danger" state, for example on the delete link.
5. The batch actions vertical-overflow-menu on small screens doesn't seem to do anything. If it's responsive, should the buttons that have been hidden appear in there?
6. I don't see JSDoc comments for the events fired by the table, `on-selected-rows-changed` and `on-page-changed`. Make sure to always document events at the top of the component file.
7. Sortable headers need correct focus styles, with the spruce outline. Maybe negative outline offset too so they don't overflow.
8. Button size issues seem to have already been addressed.
9. We'll have to figure something out for `colspan`, maybe in the next iteration.

peterpham-kyndryl:
> 1. One thing I notice is the table-toolbar is included within the overflow-x table-container, which means the table title and buttons can scroll off screen. Only the table itself should be in the overflow container. This is noticeable on the Basic Table story, but some stories seem to have nested table-containers.
>    
>    1. The "Sticky Header" story suffers here too, there seem to be multiple vertical scrollbars.
> 2. Maybe combine the batch actions from the story called "Data Table" into the Multi Select story, since those two things are interconnected.
> 3. The inline actions should support a "danger" state, for example on the delete link.
> 4. The batch actions vertical-overflow-menu on small screens doesn't seem to do anything. If it's responsive, should the buttons that have been hidden appear in there?
> 5. I don't see JSDoc comments for the events fired by the table, `on-selected-rows-changed` and `on-page-changed`. Make sure to always document events at the top of the component file.
> 6. Sortable headers need correct focus styles, with the spruce outline. Maybe negative outline offset too so they don't overflow.
> 7. Button size issues seem to have already been addressed.
> 8. We'll have to figure something out for `colspan`, maybe in the next iteration.

@brian-patrick-3: changes pushed for 1, 2, 4, 5, 6, and 7.
3. The inline actions can be customized by the devs.
8. I had it working using native <td> elements. I'm currently working on a feature where the current kyn-tr (table row) has a border-bottom, and when it's expanded, that border should disappear, and the newly added row should display its border smoothly.

brian-patrick-3:
I still have some concerns for:

3. How would a developer apply a danger state to one of these items, manual CSS? I don't think that is a friendly solution. Also using divs in the example is misleading, they will be links or buttons. IMO, the overflow menu should actually be it's own component, like Pagination.
4. It seems like this is an adaptive solution, not responsive, so developers have to add each link twice. I think it needs to be made fully responsive, or something. Also, why is the menu a table? That seems unnecessary.

peterpham-kyndryl:
> I still have some concerns for:
> 
> 3. How would a developer apply a danger state to one of these items, manual CSS? I don't think that is a friendly solution. Also using divs in the example is misleading, they will be links or buttons. IMO, the overflow menu should actually be it's own component, like Pagination.
> 4. It seems like this is an adaptive solution, not responsive, so developers have to add each link twice. I think it needs to be made fully responsive, or something. Also, why is the menu a table? That seems unnecessary.
> 
> Additionally, both menus need to close when clicking elsewhere on the page. I've done this on some of the header components, look for `handleClickOut`.

3. The content of a cell is at the discretion of the developers. The inline actions component I've included is for demonstration purposes. The column utilizes the cellRenderer function to allow developers to render the content as they see fit.
4. The appearance of table batch actions, including the overflow menu, is subject to the feature's design decisions. It can be a standalone component, providing flexibility for customization. If we aim to maintain consistency in the overflow menu, it should be developed as a separate component, such as a modal or a flyout. It's not an integral part of the table. What I've created here serves as a demonstration of how to implement actions within the data table.

brian-patrick-3 (approved):

PR #66, opened 2024-01-10
feat: modal

brian-patrick-3:


brian-patrick-3:
> Have to adjust the "X" as it shouldn't be inline with the Modal Title. Needs to be up and to the right more. The design was incorrect. Will change and update.

Updated the X position (16px from top/right), and added hover styles.

CunningDesigns (approved):
approved

PR #97, opened 2024-02-16
fix: form validation

### LINKED ISSUE #88: [Forms] Fix form validation reporting for inputs that have not been interacted with yet
**Describe the bug**

Is there a standard API exposed which would tell if the form and/or form elements valid ?

I am creating a form using all sorts of form elements like input, text area, dropdown, radio etc. I need to know if the individual form element are valid and overall form is valid state or not; no required field should be empty. So is there a standard method exposed to check for each element or form ?  Please note that it should expose whether form elements are pristine or dirty.

  Currently i do get the validity of the form using a hack by attaching listener to each element and if check the form element is required and it's not empty or not. I am sure it is not the best solution. Do we have any alternative ?

Ex:

image

brian-patrick-3:
## Summary

This resolves #88 when complete. 

We must rework each form field components to decouple the `setValdiity` state from the `internalValidationMessage`. `setValidity` must always be correct even before user input so the validity is correct when `form.submit` or `form.reportValidity()` are called, whereas `internalInvalidationMessage` must wait for user interaction or for the `invalid` event on the host.

I've completed a few components already to serve as an example. Here is the full checklist:
- [x] Text Input
- [x] Text Area
- [x] Checkbox Group
- [ ] Radio Button Group
- [ ] Dropdown
- [ ] Date Picker
- [ ] Date Range Picker
- [ ] Timepicker
- [ ] Toggle Button

CC: @kyndryl-design-system/ui-devs

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -2606,7 +2606,15 @@
             {
               "kind": "method",
               "name": "_validate",
-              "privacy": "private"
+              "privacy": "private",
+              "parameters": [
+                {
+                  "name": "interacted",
+                  "type": {
+                    "text": "Boolean"
+                  }
+                }
+              ]
             },
             {
               "kind": "method",
@@ -2661,6 +2669,19 @@
                   "name": "revealed",
                   "type": {
                     "text": "boolean"
+                  }
+                }
+              ]
+            },
+            {
+              "kind": "method",
+              "name": "_handleInvalid",
+              "privacy": "private",
+              "parameters": [
+                {
+                  "name": "e",
+                  "type": {
+                    "text": "any"
                   }
                 }
               ]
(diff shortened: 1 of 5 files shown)


brian-patrick-3:
@bikashp2023 FYI, if you'd also like to review.

src/components/reusable/datePicker/datepicker.test.ts
```diff
@@ -7,50 +7,4 @@ suite('kyn-date-picker', () => {
     const el = document.createElement('kyn-date-picker');
     assert.instanceOf(el, DatePicker);
   });
-
```
maulik-thaker1, line 10: any specific reason we have removed these test cases ?

maulik-thaker1: ok Rest is LGTM.

maulik-thaker1 (approved):

PR #105, opened 2024-02-23
fix(modal): add discrete transitions for supported browsers

brian-patrick-3:
## Summary

- Browsers that don't support discrete transitions will animate in, but not animate out. 
- Browsers that do support it will get both. 
- Once a browser adds support, it will be automatic.
- Changed transition/animation duration to 400ms.


maulik-thaker1 (approved):
Looks awesome to me till **allow-discrete** supports for browsers.

PR #118, opened 2024-03-07
fix(style): Dialog height while drawer is closing

maulik-thaker1:
## Summary

- While closing drawer it's height getting cut and then it's being closed

## Screenshots

- If you see carefully while closing drawer, firstly the height of drawer is reduced and then it's getting closed. I fixed by setting `min-height: 100vh`; to `dialog` style

https://github.com/kyndryl-design-system/shidoka-applications/assets/140634142/05c14d5b-63ae-4a63-ad7b-5a5da38f52af

### DIFF
--- a/src/components/reusable/sideDrawer/sideDrawer.scss
+++ b/src/components/reusable/sideDrawer/sideDrawer.scss
@@ -17,7 +17,7 @@ dialog {
   top: 0;
   right: 0;
   left: auto;
-  height: 100vh;
+  min-height: 100vh;
 
   &:modal {
     max-height: 100vh;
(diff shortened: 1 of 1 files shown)


src/components/reusable/sideDrawer/sideDrawer.scss
```diff
@@ -17,7 +17,7 @@ dialog {
   top: 0;
   right: 0;
   left: auto;
-  height: 100vh;
+  min-height: 100vh;
 
   &:modal {
     max-height: 100vh;
```
brian-patrick-3, line 23: I think there's actually a cleaner fix. Just move the `max-height: 100vh;` from `dialog:modal` to `dialog`. This happens because the `:modal` pseudo-selector goes away on close.

maulik-thaker1: Done.

brian-patrick-3 (approved):

PR #122, opened 2024-03-12
feat(page-title): Page title component 

maulik-thaker1:
## Summary

As per discussion with UX :

- This component is limited upto certain width according to Title text 
- Do not show Tags and action buttons. ( UX confirmation received )

**Figma link:** : 
https://www.figma.com/file/pQKkip0UrZqEbaGN2dQ3dY/Istanbul-Release?type=design&node-id=352-128746&mode=design&t=7cFpK22YMgm8fUCg-0

## ADO Story Link

https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_queries/edit/1391648/?queryId=eb0a1dfd-adb0-4453-b237-795dce5e7ea3

## Checklist

- [x] Added/updated a basic unit test, at least, for each component.


src/components/reusable/pagetitle/pageTitle.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Page Title
+ */
+
+@customElement('kyn-page-title')
+export class PageTitle extends LitElement {
+  static override styles = PageTitleScss;
+
+  /** Whether to show headline. */
+  @property({ type: Boolean })
+  showheadLine = false;
```
brian-patrick-3, line 18: These boolean "show" props might be unnecessary. Just check if the string prop is empty or not instead.

src/components/reusable/pagetitle/pageTitle.ts
```diff
@@ -0,0 +1,90 @@
...
+  @property({ type: String })
+  headLine = '';
+
+  /** Page title text (required). */
+  @property({ type: String })
+  pageTitle = '';
+
+  /** Whether to show subtitle. */
+  @property({ type: Boolean })
+  showSubTitle = false;
```
brian-patrick-3, line 30: Same comment about "show" props not being needed.

src/components/reusable/pagetitle/pageTitle.ts
```diff
@@ -0,0 +1,90 @@
...
+      <div class="page-title-wrapper">
+        ${this.showIcon
+          ? html`<div class="icon-wrapper">
+              <kd-icon .icon=${this.icon} sizeOverride="56"></kd-icon>
+            </div>`
+          : null}
+
+        <div class="page-title-text-wrapper">
+          <!-- Headline -->
+          ${this.showheadLine
```
brian-patrick-3, line 69: Can be simplified to `this.headLine !== ''`.

src/components/reusable/pagetitle/pageTitle.ts
```diff
@@ -0,0 +1,90 @@
...
+    const subTitleClasses = {
+      'page-subtitle': true,
+      [`page-subtitle-${this.type}`]: true,
+    };
+
+    return html`
+      <div class="page-title-wrapper">
+        ${this.showIcon
+          ? html`<div class="icon-wrapper">
+              <kd-icon .icon=${this.icon} sizeOverride="56"></kd-icon>
```
brian-patrick-3, line 63: Change the icon to use a `<slot>`. Remember the application may not be using vanilla carbon icons or kd-icon at all.

src/components/reusable/pagetitle/pageTitle.scss
```diff
@@ -0,0 +1,55 @@
...
+  display: block;
+}
+
+.page-title {
+  margin: 0;
+  &-wrapper {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    padding: 0 16px;
```
brian-patrick-3, line 14: This left/right padding might need to be removed. In a grid or page layout, it will un-align the title from other elements.

src/components/reusable/pagetitle/pageTitle.ts
```diff
@@ -0,0 +1,90 @@
...
+        <div class="page-title-text-wrapper">
+          <!-- Headline -->
+          ${this.showheadLine
+            ? html`<div class="page-headline">${this.headLine}</div>`
+            : null}
+          <!-- Title -->
+          <h1 class="${classMap(classes)}">${this.pageTitle}</h1>
+          <!-- Subtitle -->
+          ${this.showSubTitle
+            ? html`<h6 class="${classMap(subTitleClasses)}">
```
brian-patrick-3, line 76: Using `<h6>` here will breaking heading hierarchy. It needs to be `<h2>` if under an `<h1>`. Maybe better to not use an `<h>` tag at all for this, considering "headline" does not use one either. Generally we want the application to have full control of heading hierarchy.

src/components/reusable/pagetitle/pageTitle.scss
```diff
@@ -0,0 +1,55 @@
...
+    @include typography.type-headline-06;
+    max-width: 976px;
+  }
+}
+.page {
+  &-headline {
+    @include typography.type-body-01;
+    font-weight: 300;
+    margin-bottom: 4px;
+    max-width: 380px;
```
brian-patrick-3, line 33: Why the `max-width`s? Unless the design calls for them, just let word wrapping happen naturally.

brian-patrick-3 (changes requested):

brian-patrick-3:
Maybe also add a story that shows how the page title would look with elements such as buttons to the right of it, like in the figma file's "Title with Action Buttons" example.

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: What if they give space ?

brian-patrick-3:
IT looks like we also need a variant with white/inversed text, for use on banners/image backgrounds.

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: I also had thought in my mind but I worried about if someone give space at first character then we also need to check **this.headline !== ''** & **this.headline !== ' '** kind of

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: hmm will provide `<slot name="icon"></slot>` instead of `<kd-icon>`

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: will replace all <h> with <div> then

maulik-thaker1:
> Maybe also add a story that shows how the page title would look with elements such as buttons to the right of it, like in the figma file's "Title with Action Buttons" example.

**Action buttons** and **Tags** are removed from Page Title component. I confirmed with UX

src/components/reusable/pagetitle/pageTitle.ts
brian-patrick-3: Not sure why anyone would pass a string with a single space, but I've used this same pattern everywhere else that has optional string props without any complaints.

src/components/reusable/pagetitle/pageTitle.ts
brian-patrick-3: I think we can keep `<h1>`, as this is always intended to be the first heading at the top of a page, right?

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: ok will replace all props with this.headline !== '' this.subTitle !== '' instead of additional boolean props

brian-patrick-3:
> **Action buttons** and **Tags** are removed from Page Title component. I confirmed with UX

Right, but could they be used external from the component, inline/next to it? Just making sure. Chandan should remove those examples from Figma if not.

src/components/reusable/pagetitle/pageTitle.scss
maulik-thaker1: Its there in UX : 16px at right and left 

Screenshot 2024-03-13 at 7 33 24 PM

brian-patrick-3: I see, but that doesn't matter when it's used in practice in an actual page layout. Try it out and you'll see.

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: yes. for rest, will replace with `<div>`

maulik-thaker1:
> > **Action buttons** and **Tags** are removed from Page Title component. I confirmed with UX
> 
> Right, but could they be used external from the component, inline/next to it? Just making sure. Chandan should remove those examples from Figma if not.

yes I talked with her. She removed that **Action button** and **Tag** thing from figma

maulik-thaker1:
> IT looks like we also need a variant with white/inversed text, for use on banners/image backgrounds.

Need to talk with UX about this if we wanna do it at dev side.
I think this might be the part of theme kind of concept. 

@kyndryl-design-system/designers

brian-patrick-3:
> I think this might be the part of theme kind of concept.

Unfortunately not theme related, as the Banner example is still a white theme. The title is just used on a colorful background, so it needs the ability to invert the text color in that scenario.

If you did not specify a CSS `color` internally, it can be styled externally via inheritance at least, so we could avoid the need for a prop.

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: one thing is UX wants icon to be 56 * 56 so that's why I put directly <kd-icon> here. If we are providing slots then consumer gonna put any size icons at their end

brian-patrick-3: That's ok, we just have to provide the icon usage guidelines.

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: Done

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: Done

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: Done

src/components/reusable/pagetitle/pageTitle.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        <div class="page-title-text-wrapper">
+          <!-- Headline -->
+          ${this.showheadLine
+            ? html`<div class="page-headline">${this.headLine}</div>`
+            : null}
+          <!-- Title -->
+          <h1 class="${classMap(classes)}">${this.pageTitle}</h1>
+          <!-- Subtitle -->
+          ${this.showSubTitle
```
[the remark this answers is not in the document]
maulik-thaker1: Done

src/components/reusable/pagetitle/pageTitle.ts
maulik-thaker1: Done

brian-patrick-3 (approved):
Might need to run `analyze` one more time, but otherwise LGTM!

maulik-thaker1:
> Might need to run `analyze` one more time, but otherwise LGTM!

yup. doing same

PR #132, opened 2024-03-20
feat(card): Card component

maulik-thaker1:
## Summary

- Card components with Single slot
- Clickable and simple variant

## ADO Story 

https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/1589763

## Notes

- As per UX & UI discussion we (dev) need to provide card wrapper with single slot inside with Clickable variant.

## Checklist

- [x] Added/updated a basic unit test, at least, for each component.


src/components/reusable/card/card.ts
```diff
@@ -0,0 +1,145 @@
...
+ * @slot card-links - Slot for card links.
+ * @slot unnamed - Slot for card body content.
+ */
+
+@customElement('kyn-card')
+export class Card extends LitElement {
+  static override styles = CardScss;
+  /** Card width. Default 264px. */
+  @property({ type: Number })
+  cardWidth = 264;
```
brian-patrick-3, line 23: Width should be dictated by the Grid, so the card component itself should always be 100% width.

src/components/reusable/card/card.ts
```diff
@@ -0,0 +1,145 @@
...
+   * @ignore
+   */
+  @state()
+  isCollapse = false;
+
+  override render() {
+    return html`${this.type === 'clickable'
+      ? html`<a
+          href=${this.cardLink}
+          target="_blank"
```
brian-patrick-3, line 64: We shouldn't force new tab here. We also will need to add an event to capture link clicks.

src/components/reusable/card/card.scss
```diff
@@ -0,0 +1,105 @@
...
+  padding: 24px;
+  background-color: var(--kd-color-background-ui-default);
+  border: 1px solid var(--kd-color-border-light);
+  border-radius: 8px;
+  //   box-shadow: 0px 1px 2px 0px rgba(61, 60, 60, 0.25);
+  outline: 1px solid transparent;
+  transition: outline 200ms ease-out;
+  &:hover {
+    outline: 2px solid var(--kd-color-border-ui-hover);
+  }
```
brian-patrick-3, line 21: Non-clickable card should not have a hover effect.

brian-patrick-3 (commented):
Added a couple initial comments

src/components/reusable/card/card.ts
brian-patrick-3: I think maybe instead of a width prop, set a CSS width on the `:host` selector. Then it can be overridden this way:
```css
kyn-card { width: 100%; }
```
It needs to be able to use percentage width, especially within a grid.

src/components/reusable/card/card.ts
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Card.
+ * @slot card-logo - Slot for card logo (Required). Size 32 * 32 as per UX guidelines.
+ * @slot card-action-button - Slot for Card action button.
+ * @slot card-overflow-menu - Slot for overflow menu.
+ * @slot card-thumbnail - Slot for card thumbnail image.
+ * @slot tags - Slot for the tag component.
+ * @slot card-links - Slot for card links.
+ * @slot unnamed - Slot for card body content.
```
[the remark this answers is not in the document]
maulik-thaker1: Done

brian-patrick-3 (changes requested):
Here's a few initial observations:

1. Inside Grid story:
   1. `md` grid col size seems incorrect, doesn't fill the screen.
   2. Show cards with varying content, so we can see that the cards have matching heights.
   3. Set `height: 100%;` on `kyn-card` so they match in height.
   4. Set `height: 100%;` on `.card-wrapper` so it grows with the :host container.
2. The 1px `border` doesn't persist between "clickable" and "normal" types, so the card size changes and content shifts slightly when changing type.
3. Based on what we just discussed about notifications, we might want to make the `.card-wrapper` div customizable, so we can alter the shadow and background color. The easiest way would be to add a CSS `part` on this tag. The downside of `part` is it opens it up to any CSS customization.
4. I wonder if we could do anything to internalize some of the pattern CSS so it's easier for consuming applications to re-use. That might require sub-components though hmm. Just some food for thought.

maulik-thaker1:
> Here's a few initial observations:
> 
> 1. Inside Grid story:
>    
>    1. `md` grid col size seems incorrect, doesn't fill the screen.
>    2. Show cards with varying content, so we can see that the cards have matching heights.
>    3. Set `height: 100%;` on `kyn-card` so they match in height.
>    4. Set `height: 100%;` on `.card-wrapper` so it grows with the :host container.
> 2. The 1px `border` doesn't persist between "clickable" and "normal" types, so the card size changes and content shifts slightly when changing type.
> 3. Based on what we just discussed about notifications, we might want to make the `.card-wrapper` div customizable, so we can alter the shadow and background color. The easiest way would be to add a CSS `part` on this tag. The downside of `part` is it opens it up to any CSS customization.
> 4. I wonder if we could do anything to internalize some of the pattern CSS so it's easier for consuming applications to re-use. That might require sub-components though hmm. Just some food for thought.

1. Inside Grid : 
    i. I changed and did this : class="kd-grid__col--sm-2 kd-grid__col--md-4 kd-grid__col--lg-3"
    ii. ok for this I need to set props inside `<sample-card-component>` right now I hard-coded Title, Subtitle and Description  inside `<sample-card-component>`
    iii. Set. it's already matching actual height of card
    iv. ok
2. yeah its `border` for `normal` and `outline` for `clickable`. I changed that and made `outline` for both
3. Need to think about that as right now as per UX there is white background for card and box-shadow for clickable card

brian-patrick-3:
In the grid example, `kyn-card` still needs `height: 100%;` so they are all the same height.

src/components/reusable/card/card.scss
```diff
@@ -0,0 +1,34 @@
...
+  &-wrapper-clickable {
+    position: relative;
+    display: inline-block;
+    color: unset;
+    width: inherit;
+    text-decoration: none;
+    background-color: var(--kd-color-background-ui-default);
+    outline: 1px solid var(--kd-color-border-ui);
+    border-radius: 8px;
+    box-shadow: 0px 1px 2px 0px rgba(61, 60, 60, 0.25);
```
brian-patrick-3, line 27: Pull in [Elevation](https://kyndryl-design-system.github.io/shidoka-foundation/?path=/docs/foundation-elevation--docs) here, now that it's available.

src/components/reusable/card/card.scss
```diff
@@ -0,0 +1,34 @@
...
+    padding: 24px;
+    transition: outline 200ms ease-out;
+    &:hover {
+      outline: 2px solid var(--kd-color-border-ui-hover);
+    }
```
brian-patrick-3, line 32: Hover needs Elevation level 2

maulik-thaker1: ok will use elevation here. Transition is there : ` transition: outline 200ms ease-out;`

src/components/reusable/card/card.scss
maulik-thaker1: Can you give me an example how to use that ?

src/components/reusable/card/card.scss
maulik-thaker1: Hover need elevation level 1 only as per figma for clickable version.

brian-patrick-3 (approved):

PR #133, opened 2024-03-20
feat(local-nav): local nav redesign

brian-patrick-3:


maulik-thaker1:
It is opening like this on hover:
Screenshot 2024-03-22 at 11 35 36 PM

maulik-thaker1 (approved):
Just animation lag I observed when you quickly hovered on sub menus 
otherwise LGTM.

PR #137, opened 2024-03-28
feat(table): refactor table component for conditional rendering

peterpham-kyndryl:
## Summary

Link here (if applicable).

## Notes

- Detailed change notes
- can go here

## To Do

- Anything else
- left to do

## Checklist

## Testing Instructions

## Screenshots

(if any)


package.json
```diff
@@ -29,6 +29,7 @@
   "dependencies": {
     "@carbon/icon-helpers": "^10.37.0",
     "@carbon/icons": "^11.14.0",
+    "@lit/context": "^1.1.0",
```
brian-patrick-3, line 32: This is neat, I didn't even know it was a thing. Could be very useful for component > sub-component communication.

brian-patrick-3 (changes requested):
This is looking pretty great, thanks for all your hard work!

Here's a few things I think we still need:

1. Try to restore the deleted components and stories, and avoid any breaking changes if possible as we discussed after the demo. This is what I'm most worried about.
2. Add a Story that shows the table with Pagination. People will be looking for that.
3. The example code is hard to find with the Storybook changes. I would suggest adding links to the sample code/component file on GitHub somewhere in the Docs.
4. We'll also need a Story in the Global Filter that shows it working with a Table. Can maybe follow up on that after this release.

peterpham-kyndryl:
> This is looking pretty great, thanks for all your hard work!
> 
> Issues I see:
> 
> 1. Selectable version - row checkboxes are not aligned with header checkbox.
> 
> Here's a few things I think we still need:
> 
> 1. Try to restore the deleted components and stories, and avoid any breaking changes if possible as we discussed after the demo. This is what I'm most worried about.
>    
>    1. If breaking changes are unavoidable:
>       
>       1. Make sure your [commit messages call out each breaking change](https://www.conventionalcommits.org/en/v1.0.0/#summary).
>       2. Point this PR to `beta` branch, as we'll have to start a 2.0 release.
> 2. Add a Story that shows the table with Pagination. People will be looking for that.
> 3. The example code is hard to find with the Storybook changes. I would suggest adding links to the sample code/component file on GitHub somewhere in the Docs.
> 4. We'll also need a Story in the Global Filter that shows it working with a Table. Can maybe follow up on that after this release.

@brian-patrick-3 
Thank you!
Regarding the checkboxes issue - I wasn't aware of it since it only happened on FireFox but it's fixed now. I tested on all browsers (chrome, safari on Mac, chrome, firefox and edge), and it's currently working fine on all of them.

Regarding things that we still need:
1. Breaking changes: I double-checked and there's no breaking change. They can keep using the deprecated data table and old basic table. Old stories are now restored with just a tiny change.
2. Pagination stories added to the new table.
3. Link added. Later I can provide my demos in React, Svelte (and probably Angular) so it's easier for them to follow.
4. Sure thing!

brian-patrick-3 (approved):

Issue #37, opened 2023-11-30
Kyn Dropdown - Pre Selected Value from List Not Coming Up


guptaashray:
This is the list I'm populating in the single select dropdown - 
[
    {
        "name": "Highly Critical",
        "content": "Highly Critical",
        "selected": false
    },
    {
        "name": "Somewhat critical",
        "content": "Somewhat critical",
        "selected": false
    },
    {
        "name": "Low",
        "content": "Low",
        "selected": true
    }
]

```
<kyn-dropdown [searchable]="false" size="xl" [placeholder]="selectCriticality" [disabled]="!service.businessCriticality.length" (on-change)="onInputDefnValuesCriticality($event)">
          <kyn-dropdown-option *ngFor="let option of businessCriticalitites" [value]="option.content [selected]="option.selected">{{option['content']}}</kyn-dropdown-option>
</kyn-dropdown>
```
The above is my HTML code for the same.

Now, the above values are rendered like this where I can see the selected value only when I open the dropdown in the second pic,  but does not appear when I don't open the dropdown as in the first pic.

<img width="838" alt="Screenshot 2023-11-30 at 10 42 04 PM" src="https://github.com/kyndryl-design-system/shidoka-applications/assets/43576436/0093fb4c-6dfb-4988-aa33-9b16755c8305">

<img width="805" alt="Screenshot 2023-11-30 at 10 42 31 PM" src="https://github.com/kyndryl-design-system/shidoka-applications/assets/43576436/46210bba-afb8-40b0-81dc-ff04232902fa">

Now if I open the dropdown and select any other option, it will appear.

<img width="796" alt="Screenshot 2023-11-30 at 10 45 19 PM" src="https://github.com/kyndryl-design-system/shidoka-applications/assets/43576436/3c8adf89-1ffa-4491-a95a-f6f4b3aeae19">

But, if I select the same option that was already selected, it will get deselected, but then if I again try to select that same option, I'm unable to see it appearing when the dropdown closes.
To see the already selected option appear, I need to open the dropdown and select any other option, and again select that option.

Expected - The already selected option should appear in the dropdown and the user should be able to deselect and select back the same option without selecting any other option.

https://github.com/kyndryl-design-system/shidoka-applications/assets/43576436/967cd69a-a844-45b3-ba5e-1436ca59bf3f

brian-patrick-3:
[...]

PR #142, opened 2024-04-04
feat(notification): Read / unread

maulik-thaker1:
## Summary

- Add state of read and unread inside Notification panel and Notification
- Change in css of kd-card for background changes according to read / unread state of notification


src/components/global/header/headerNotificationPanel.ts
```diff
@@ -35,6 +57,13 @@ export class HeaderNotificationPanel extends LitElement {
             <h1 class="panel-header-text">${this.panelTitle}</h1>
           </div>
           <div class="panel-left-slot">
+            <kd-button
+              kind="tertiary"
+              @click=${(e: Event) => this._toggleMarkAllReadUnread(e)}
+              >${this.markAllRead
+                ? this.textStrings.markAllUnread
+                : this.textStrings.markAllRead}</kd-button
+            >
```
brian-patrick-3, line 66: I think we should externalize this, and place it in the menu-slot with the settings icon.

src/components/reusable/notification/notification.scss
```diff
@@ -5,9 +5,9 @@
 :host {
   display: block;
   width: 100%;
-  kd-card {
+  kyn-card {
     width: 100%;
-    margin-top: 8px;
+    padding-bottom: 8px;
```
brian-patrick-3, line 10: Don't put padding/margin here. Use CSS in the notifcation-panel component to create the gaps.

maulik-thaker1:
Need [#37](https://github.com/kyndryl-design-system/shidoka-foundation/pull/37) merge first.

maulik-thaker1:
> Need [#37](https://github.com/kyndryl-design-system/shidoka-foundation/pull/37) merge first.

done and bumped foundation version [1.6.0 ](https://github.com/kyndryl-design-system/shidoka-foundation/releases/tag/v1.6.0)

src/components/reusable/card/card.ts
```diff
@@ -51,7 +52,7 @@ export class Card extends LitElement {
         >
           <slot></slot>
         </a>`
-      : html`<div class="card-wrapper">
+      : html`<div part="card-wrapper" class="card-wrapper">
```
brian-patrick-3, line 55: You don't have to update this file anymore since it's deprecated.

src/components/reusable/notification/notification.scss
brian-patrick-3: This padding is still here.

src/components/reusable/notification/notification.ts
```diff
@@ -60,10 +59,17 @@ export class Notification extends LitElement {
     error: 'failed',
   };
 
+  /** Set notification mark read prop.*/
+  @property({ type: Boolean, reflect: true })
+  markRead: any = false;
```
brian-patrick-3, line 64: Right now, "unread" is the default state, is that right? Might want to confirm that with Marco, because read/unread functionality might not be used in every case. In the case there is no unread/read functionality, I would expect "read" styles to be the default.

So what I propose is naming this prop "unread", and default = false.

brian-patrick-3 (changes requested):

src/components/reusable/notification/notification.ts
maulik-thaker1: yes **unread** is default state. I am changing this prop name as **unread**

maulik-thaker1: done

brian-patrick-3 (changes requested):
Can you enable the "unRead" control on the Notification story? Besides that and the built-in padding, I think this is ready.

src/components/reusable/notification/notification.scss
maulik-thaker1: oh ok removing this

maulik-thaker1: here you go. Its done and removed from here.

brian-patrick-3 (approved):

maulik-thaker1:
> Can you enable the "unRead" control on the Notification story? Besides that and the built-in padding, I think this is ready.

Done. Added `unRead` prop for notification story as well

PR #149, opened 2024-04-16
feat(widget): add widget component

brian-patrick-3:


maulik-thaker1:
Curious about this `disabled` functionality. Here only background is lighten but we can still able to click widget internals when `disabled = true`. Is it as needed?

Uploading Screen Recording 2024-04-26 at 5.01.59 PM.mov…

src/components/reusable/widget/widget.ts
```diff
@@ -0,0 +1,91 @@
...
+  @property({ type: String })
+  widgetTitle = '';
+
+  /** Widget sub-title. */
+  @property({ type: String })
+  subTitle = '';
+
+  /** Widget drag active state. */
+  @property({ type: Boolean })
+  dragActive = false;
```
maulik-thaker1, line 27: What else `dragActive` do apart from adding shadow ?

brian-patrick-3: That's it.

brian-patrick-3:
> Curious about this `disabled` functionality. Here only background is lighten but we can still able to click widget internals when `disabled = true`. Is it as needed?

Yeah makes sense because we don't have control over the slotted content.

PR #154, opened 2024-04-22
fix(table): used the default values from the sub components instead o…

peterpham-kyndryl:
…f the table. Only apply new values if they are set. Fix multi selection when pre-checked values are passed in

## Summary

Link here (if applicable).

## Notes

- Detailed change notes
- can go here

## To Do

- Anything else
- left to do

## Checklist

## Testing Instructions

## Screenshots

(if any)

### DIFF
--- a/src/components/reusable/table/table-header.ts
+++ b/src/components/reusable/table/table-header.ts
@@ -1,4 +1,4 @@
-import { html, LitElement } from 'lit';
+import { html, LitElement, PropertyValues} from 'lit';
 import {
   customElement,
   property,
@@ -29,6 +29,28 @@ import { SORT_DIRECTION, TABLE_CELL_ALIGN } from './defs';
 @customElement('kyn-th')
 export class TableHeader extends LitElement {
   static override styles = [styles];
+
+  /**
+   * Sets a fixed width for the cell.
+   * Accepts standard CSS width values (e.g., '150px', '50%').
+   */
+  @property({ type: String, reflect: true })
+  width = '';
+
+  /**
+   * Sets a maximum width for the cell; contents exceeding this limit will be truncated with ellipsis.
+   * Accepts standard CSS width values (e.g., '150px', '50%').
+   */
+  @property({ type: String, reflect: true })
+  maxWidth = '';
+
+  /**
+   * Sets a minimum width for the cell;
+   * Accepts standard CSS width values (e.g., '150px', '50%').
+   * @type {string}
+   */
+  @property({ type: String, reflect: true })
+  minWidth = '';
 
   /** Determines if the cell should have a denser layout. */
   @property({ type: Boolean, reflect: true })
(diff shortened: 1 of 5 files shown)


maulik-thaker1 (changes requested):
Here its shows **boolean** & **undefined** it should be **boolean** IMO

Screenshot 2024-04-23 at 5 57 49 PM

maulik-thaker1:
Here as well. Need to do `npm run analyze` once

Screenshot 2024-04-23 at 6 01 42 PM

peterpham-kyndryl:
> Here its shows **boolean** & **undefined** it should be **boolean** IMO
> 
> Screenshot 2024-04-23 at 5 57 49 PM

Yes, they're boolean, but they are optional. Looks like Storybook auto docs is not generating the optional props properly. I wanted these props to be interpreted as that these are optional, but if they're passed in they have to be boolean. . 
Anyways, I agree that this has to be fixed.

src/components/reusable/table/table-header.ts
```diff
@@ -157,8 +179,21 @@ export class TableHeader extends LitElement {
...
 
-  override updated() {
+  override updated(changedProperties: PropertyValues) {
     this.getTextContent();
+
+    super.updated(changedProperties);
+    if (this.maxWidth && changedProperties.has('maxWidth')) {
+      this.style.setProperty('--kyn-th-max-width', this.maxWidth);
+    }
+
```
maulik-thaker1, line 189: IMO no need to check `this.maxWidth` here as it's already covered in `changedProperties.has('maxWidth')`. So I guess your if condition will be 

`if(changedProperties.has('maxWidth'){ ... }`

Same applies for other if conditions also. [`this.minWidth` and `this.width`]

maulik-thaker1 (changes requested):

src/components/reusable/table/table.ts
```diff
@@ -25,35 +25,35 @@ export class Table extends LitElement {
    * selectable using checkboxes.
    */
   @property({ type: Boolean })
-  checkboxSelection = false;
+  checkboxSelection?: boolean;
```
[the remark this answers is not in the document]
peterpham-kyndryl: Like I mentioned previously, these are optional so the default values will be undefined. When they're not set, the sub-components will use their own default values. Those values will only be overridden if these values are really set by the users.
The way updated function works is that even if the user doesn't set a new value, it's still invoked at the first time when the values are set from undefined to the default values. And this is not what I want to happen, so I need them to be undefined by default, not false.

src/components/reusable/table/table-header.ts
peterpham-kyndryl: changedProperties.has('maxWidth') only checks if there is a change of maxWidth, meanwhile this.maxWidth is checking for a truthy (i.e., it has a value other than undefined, null, 0, NaN, an empty string, or false). If this.width is falsy, the property is not set, which can prevent potential issues or errors.

peterpham-kyndryl:
Closing since changes are included in this PR https://github.com/kyndryl-design-system/shidoka-applications/pull/157

PR #157, opened 2024-04-25
feat(table): added new feature - expanded and disabled rows

peterpham-kyndryl:
## Summary

Link here (if applicable).

## Notes

- Detailed change notes
- can go here

## To Do

- Anything else
- left to do

## Checklist

## Testing Instructions

## Screenshots

(if any)

### DIFF
--- a/src/components/reusable/table/table-expanded-row.scss
+++ b/src/components/reusable/table/table-expanded-row.scss
@@ -1,0 +1,27 @@
+:host {
+  display: table-row;
+  block-size: 0;
+
+  td {
+    padding: 0;
+    vertical-align: middle;
+
+    .child-row-inner-container {
+      overflow: hidden;
+      block-size: 0;
+    }
+  }
+}
+
+:host([expanded]) {
+  block-size: auto;
+
+  td {
+    block-size: auto;
+    border-bottom: 1px solid var(--kd-color-border-light);
+
+    .child-row-inner-container {
+      block-size: auto;
+    }
+  }
+}
--- a/src/components/reusable/table/table-expanded-row.ts
+++ b/src/components/reusable/table/table-expanded-row.ts
@@ -1,0 +1,50 @@
+import { LitElement, html } from 'lit';
+import { property, customElement } from 'lit/decorators.js';
+
+import styles from './table-expanded-row.scss';
+
+/**
+ *
+ * `kyn-expanded-tr` Web Component.
+
+ * Designed to display additional details for a row in a table.
+ * The row is expandable and can be expanded/collapsed by toggling the plus/minus icons.
+ *
+ * @slot unnamed - The slot for adding content to the expandable details section.
+ */
+@customElement('kyn-expanded-tr')
+export class TableExpandedRow extends LitElement {
+  static override styles = [styles];
+
+  /**
+   * The number of columns that the expanded row should span.
+   * Reflects the `colspan` attribute.
+   */
+  @property({ type: Number, attribute: 'colspan' })
+  colSpan = 1;
+
+  /**
+   * `true` if the table row should be expanded.
+   */
+  @property({ type: Boolean, reflect: true })
+  expanded = false;
+
+  override render() {
+    const { colSpan } = this;
+    return html`
+      <td></td>
+      <td colspan="${colSpan}">
+        <div class="child-row-inner-container">
+          <slot></slot>
+        </div>
+      </td>
+    `;
+  }
+}
+
+// Define the custom element in the global namespace
+declare global {
+  interface HTMLElementTagNameMap {
+    'kyn-expanded-tr': TableExpandedRow;
+  }
+}
--- a/src/components/reusable/table/table.ts
+++ b/src/components/reusable/table/table.ts
@@ -23,46 +23,58 @@ export class Table extends LitElement {
   /**
    * checkboxSelection: Boolean indicating whether rows should be
    * selectable using checkboxes.
+   * @type {boolean}
+   * @default false
    */
   @property({ type: Boolean })
-  checkboxSelection = false;
+  checkboxSelection?: boolean;
 
   /**
    * striped: Boolean indicating whether rows should have alternate
    * coloring.
+   * @type {boolean}
+   * @default false
    */
   @property({ type: Boolean })
-  striped = false;
+  striped?: boolean;
 
   /**
    * stickyHeader: Boolean indicating whether the table header
    * should be sticky.
+   * @type {boolean}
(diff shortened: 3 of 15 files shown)


maulik-thaker1:
For disable row its hard to understand only checkbox look like is disabled. I don't know the UX but at dev end cursor should be **not-allowed** for whole row and row should look separate out when disabled. Have you talked this with Robert ?
Screenshot 2024-04-25 at 5 55 23 PM

maulik-thaker1 (approved):
Overall LGTM. Left few comments for `disabled` version. Check with UX for this. 
Also, verify once with Robert.

peterpham-kyndryl:
> Overall LGTM. Left few comments for `disabled` version. Check with UX for this. Also, verify once with Robert.

Nice catch, Thanks Maulik!
I somehow reverted the text color part. Text should look disabled too. Besides that, the background color doesn't change (confirmed with Robert), and since all actions are not allowed, "pointer-events: none" is used instead of "cursor: not-allowed"

maulik-thaker1:
> > Overall LGTM. Left few comments for `disabled` version. Check with UX for this. Also, verify once with Robert.
> 
> Nice catch, Thanks Maulik! I somehow reverted the text color part. Text should look disabled too. Besides that, the background color doesn't change (confirmed with Robert), and since all actions are not allowed, "pointer-events: none" is used instead of "cursor: not-allowed"

ok.

src/components/reusable/table/table-expanded-row.scss
```diff
@@ -0,0 +1,28 @@
+:host {
+  display: table-row;
+  block-size: 0;
+  transition: height 150ms ease-out;
+
```
maulik-thaker1, line 5: Generally we are using 400ms for smooth animation. You can also check if its add more smoothing in animation.

brian-patrick-3: The transition doesn't actually do anything, since no explicit height is being set. We don't want to animate height either as it's a property that causes a layout change. Can just remove it.

src/components/reusable/table/table-expanded-row.ts
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * `true` if the table row should be expanded.
+   */
+  @property({ type: Boolean, reflect: true })
+  expanded = false;
+
+  override render() {
+    const { colSpan } = this;
+    return html`
+      <td></td>
```
brian-patrick-3, line 35: Let's just remove this empty `<td>`, have the colspan fill the entire row, and instead give some padding to the entire expanded row container. Right now the nested table blends into the next row of the parent table which is confusing.

src/components/reusable/table/table.ts
```diff
@@ -23,46 +23,58 @@ export class Table extends LitElement {
   /**
    * checkboxSelection: Boolean indicating whether rows should be
    * selectable using checkboxes.
+   * @type {boolean}
+   * @default false
    */
   @property({ type: Boolean })
-  checkboxSelection = false;
+  checkboxSelection?: boolean;
```
brian-patrick-3, line 30: Boolean props should always default to false. Is there a specific reason you might need them to remain undefined?

peterpham-kyndryl: The updated function is triggered initially when values transition from undefined to their default settings, even if these values aren't explicitly set by the user. To avoid this, I require the default values to remain undefined rather than being automatically set to false.

src/components/reusable/table/table-expanded-row.ts
brian-patrick-3: The parent table columns will never align with the child table columns because they are standalone, and need to be treated as such. I've mentioned this several times to make sure it was clear to the designers.

src/components/reusable/table/table.ts
brian-patrick-3: You can avoid this instead by checking:
```js 
if (changedProps.has('prop') && changedProps.get('prop') !== undefined) {
  // logic
}
```

peterpham-kyndryl: Tried a few solutions including this. It didn't work. When true is passed over making this.prop is true but changedProps.get('prop') at that time has a value of undefined making the condition not matched.

src/components/reusable/table/table-expanded-row.ts
peterpham-kyndryl: Actually, they can be aligned. 
Anyways, I spoke with Robert, Marco, and Matthew, they're all ok with this empty td for now. Since it looks indented with the space of the expanding column which is nice enough. They'll look for more feedbacks and explore more things regarding table UX design and what should be the padding, (fixed, responisive...etc), and what values are good.
Also, apps have the full control over the rest of the space.

src/components/reusable/table/table-expanded-row.scss
peterpham-kyndryl: I can dynamically set the height using JS.  So if we're not animating height. Do you have any suggestions?

src/components/reusable/table/table-expanded-row.ts
peterpham-kyndryl: Well, they can be aligned.
Anyways, I spoke with Matthew, Marco, and Robert about this indentation and this empty td. We'll all agree to leave it there. . UX is still gathering feedbacks and exploring things regarding table. A fixed padding-left of 64px is another option but it's fixed so not for now. Besides, changing this is not a breaking change by any means.

src/components/reusable/table/table-expanded-row.scss
brian-patrick-3: You won't necessarily know the content height. The content could be loaded asynchronously. I would say just have no open/close animation, same as Accordion.

src/components/reusable/table/table-expanded-row.ts
brian-patrick-3: Curious, how are they aligned? The only thing I can think of is fixed width columns.

brian-patrick-3 (approved):

src/components/reusable/table/table-expanded-row.ts
peterpham-kyndryl: Yes, fixed widths.

src/components/reusable/table/table-expanded-row.scss
peterpham-kyndryl: Ok. Discussed with Matthew that we'll have no animation for row exapdning.

brian-patrick-3 (approved):

PR #162, opened 2024-04-29
feat(toast-inline): Toast and Inline variants of notification

maulik-thaker1:
## Summary

- Added `toast` and `inline` variants of notification

## ADO Stories

https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/1670772
https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/1670765

## Notes

- Added `inline` and `toast` variant for  existing`<kyn-notification>` component


src/components/reusable/notification/notification.ts
```diff
@@ -119,6 +175,10 @@ export class Notification extends LitElement {
           <div class="timestamp-text">${this.timeStamp}</div>
         </div>
       </div>
+
+      <div class="notification-inline-toast-action">
+        <slot name="action-link"></slot>
```
brian-patrick-3, line 180: I don't think we need a separate slot for this link. It can just go in the unnamed slot.

src/components/reusable/notification/notification.scss
```diff
@@ -47,13 +70,53 @@ kd-card.notification-mark-read::part(card-wrapper) {
...
+.notification-toast .notification-content-wrapper {
+  padding: 0;
+}
 .notification-description ::slotted(*) {
   @include typography.type-ui-02;
   color: var(--kd-color-text-primary);
   padding-top: 10px;
 }
+.notification-inline .notification-description ::slotted(*),
+.notification-toast .notification-description ::slotted(*) {
```
brian-patrick-3, line 94: I don't really like the use of `::slotted(*)` here, it is too broad and applies to any slotted element. Can we remove all of these * rules?

src/components/reusable/notification/notification.stories.js
```diff
@@ -12,12 +17,17 @@ const handleOverflowClick = (e) => {
...
+  // To close notification
+};
+
 export default {
   title: 'Components/Notification',
   component: 'kyn-notification',
   argTypes: {
     type: {
-      options: ['normal', 'clickable'],
+      options: ['normal', 'clickable', 'inline', 'toast'],
```
brian-patrick-3, line 30: Can't inline and toast also be clickable? What if we separate "clickable" from "type" and just infer it automatically if an `href` is provided.

brian-patrick-3 (changes requested):
Had a few comments and concerns:

1. Should the Toast notification be floating on the screen? Ask Marco, but I would expect a position: fixed, placement in the corner of the screen, and possible for it to close automatically after X seconds.
2. Most of the Controls are not functional in the Inline and Toast stories.

src/components/reusable/notification/notification.ts
maulik-thaker1: okay

src/components/reusable/notification/notification.scss
maulik-thaker1: Then how to style this slotted part for description. As we added inline and toast inside <kyn-notification> so I remain UI as it is and handled things by styles. Will try to reduce if possible but can't remove all of these

src/components/reusable/notification/notification.stories.js
maulik-thaker1: Nope inline and toast can't be clickable. Will cross check with Marco once.

src/components/reusable/notification/notification.scss
brian-patrick-3: It depends what you're trying to accomplish, but you shouldn't have to style any slotted elements. Styling an internal element that wraps the slot should be sufficient.

maulik-thaker1:
> Had a few comments and concerns:
> 
> 1. Should the Toast notification be floating on the screen? Ask Marco, but I would expect a position: fixed, placement in the corner of the screen, and possibly for it to close automatically after X seconds.
> 2. Most of the Controls are not functional in the Inline and Toast stories.

1. Didn't understand this. IMO User can apply styles on it and do hide show on their own I guess. Also, they can give whatever time they want till appear

2. Yeah most of them are not useful.

src/components/reusable/notification/notification.scss
maulik-thaker1: ok will change that logic of scss.

src/components/reusable/notification/notification.stories.js
maulik-thaker1: I mean here we already used `type="clickable"` inside notification panel
Screenshot 2024-04-29 at 7 16 00 PM

brian-patrick-3: Right, but it becomes irrelevant if you just check `if (this.href !== '')` or something instead

src/components/reusable/notification/notification.ts
maulik-thaker1: Done

src/components/reusable/notification/notification.scss
maulik-thaker1: Done Removed `::slotted(*)`

marcosebastianlopez (commented):
I think the overflow is hidden on the container of the notifications themselves. Which is cutting off the sides

marcosebastianlopez:
image

marcosebastianlopez:
We also need to figure out the close button. I know these are components, but they should be more custom to be as close to the upper-right corner as possible.

image

image

maulik-thaker1:
> We also need to figure out the close button. I know these are components, but they should be more custom to be as close to the upper-right corner as possible.
> 
> image image

its a `<kd-button type="tertiary">`

brian-patrick-3:
> image

Adding enough padding on "notification-container" should resolve that.

maulik-thaker1:
> > image
> 
> Adding enough padding on "notification-container" should resolve that.

when this is appear ? How much padding should we give here?

maulik-thaker1:
> > its a `<kd-button type="tertiary">` Do you want this button to upper right corner ?
> 
> If the X is intended to be a dark/black color, you can't use tertiary button which is spruce color.

which kind do we need to use for icon only button ? Tertiary only gives no outline / border. 
https://kyndryl-design-system.github.io/shidoka-foundation/?path=/story/components-button--icon-only&args=kind:tertiary

brian-patrick-3:
> i given icon fill : #3d3c3c

Oh that works? Sure that's fine then.

maulik-thaker1:
> > when this is appear ? How much padding should we give here?
> 
> Enough padding so none of the shadows are cut off

Done

brian-patrick-3 (approved):

PR #166, opened 2024-04-30
feat(widget-gridstack): gridstack wrapper component

brian-patrick-3:


src/common/helpers/gridstack.ts
```diff
@@ -0,0 +1,76 @@
+export const Config = {
+  handle: 'kyn-widget-drag-handle', // set the drag handle
```
peterpham-kyndryl, line 2: It's my two cents that this one may be better using a class name or something else. Since later, let's say Bridge, will be able to import and consume an external widget that is not Shidoka's to the dashboard.

brian-patrick-3: At the very least, the default config can be overridden.

src/common/scss/variables/zindex.scss
```diff
@@ -0,0 +1,5 @@
+:root {
+  --kd-z-overlay: 100000;
+  --kd-z-header: 10000;
+  --kd-z-local-nav: 9000;
+}
```
maulik-thaker1, line 5: is this specific for **gridStack** or we can use this across the other components too?

brian-patrick-3: Ah yes I forgot about this. I wanted to create a z-index table of contents basically to help us choose values and prevent layering issues.

marcosebastianlopez (approved):
This is awesome! Thank you

PR #168, opened 2024-05-01
fix: flyout interactions

brian-patrick-kyndryl:
- Add Hover Intent to flyouts (150ms)
- Reduce overlay blur to 1px
- Restore Local Nav Pinned feature


marcosebastianlopez (approved):
Looks great!

PR #178, opened 2024-05-22
feat(seach): add search component

brian-patrick-3:


maulik-thaker1:
Seems auto suggest not working by selecting element from mouse click on that item list. But its working on press **Enter** key only. (Don't know if there is a requirement as such, but just observed)

https://github.com/kyndryl-design-system/shidoka-applications/assets/140634142/2b017d2f-c4f9-40e6-a964-8381696f9286

maulik-thaker1:
Expand / collapse animation is too fast. It could be smooth.

https://github.com/kyndryl-design-system/shidoka-applications/assets/140634142/53cf560b-0891-4403-9c4d-8cf09fa3919e

src/components/reusable/search/search.ts
```diff
@@ -0,0 +1,245 @@
...
+    const event = new CustomEvent('on-input', {
+      detail: Detail,
+    });
+    this.dispatchEvent(event);
+  }
+
+  private _handleSuggestionClick(suggestion: string) {
+    this.value = suggestion;
+  }
+
```
maulik-thaker1, line 167: Not working when click on any suggestion.  (Only works using Enter key pressed)

maulik-thaker1 (changes requested):
Overall LGTM. Just few observations noticed and mentioned in comments.

maulik-thaker1:
Optional text is missing according to UX

Screenshot 2024-05-24 at 6 29 43 PM

maulik-thaker1:
There is a provision to move icon **left** and **right** side and **not to show** as per UX. Are we not including this ?
( UX confirmation required )

Screenshot 2024-05-24 at 6 30 50 PM

maulik-thaker1:
Also there are various states of search like **error**, **warning**, etc. Please confirm with UX once. I am referring [this](https://www.figma.com/design/6AovH7Iay9Y7BkpoL5975s/Applications-Specs-for-Devs?node-id=5115-29153&m=dev) figma file

brian-patrick-3:
1. Fixed suggestion click. The blur function broke it, so needed to add a delay.
2. Already spoke with Robert about the transition from collapsed to expanded. There's no way to nicely animate from a button to an input, so we're keeping a hard transition for now.
3. There are several properties from text-input under the hood that I am not including in search since it is designed in a more specific way. I am also not including form/validation properties as this is not intended for use in forms, text input can be used instead.

maulik-thaker1 (approved):

PR #190, opened 2024-06-06
fix(table-setting): hide column checkbox and lock icon when scroll

maulik-thaker1:
## Summary
- Raised by Jaideep M V when we scroll the row then it overlaps the header and shows lock icon and checkbox. 
- Resolved by setting z-index

### DIFF
--- a/src/components/reusable/table/story-helpers/column-setting.ts
+++ b/src/components/reusable/table/story-helpers/column-setting.ts
@@ -46,12 +46,14 @@ class StoryColumSetting extends LitElement {
     }
     .t-head {
       top: 72px;
+      z-index: 2;
     }
     .seperator-div {
       height: 8px;
       background: white;
       position: sticky;
       top: 64px;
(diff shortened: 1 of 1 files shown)


src/components/reusable/table/story-helpers/column-setting.ts
```diff
@@ -46,12 +46,14 @@ class StoryColumSetting extends LitElement {
     }
     .t-head {
       top: 72px;
+      z-index: 2;
```
brian-patrick-3, line 49: This is not needed, as I already added a z-index to the `stickyheader`.

brian-patrick-3 (approved):

src/components/reusable/table/story-helpers/column-setting.ts
maulik-thaker1: I saw but its not working for lock icon. lock icon still visible while hover on row and scroll.

PR #208, opened 2024-07-17
fix(data-table): Added hover on table row according to UX

maulik-thaker1:
## Summary

UX added hover / rollout color when table is not interactive

## ADO Story 

https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/1808774

### DIFF
--- a/src/components/reusable/table/table-row.scss
+++ b/src/components/reusable/table/table-row.scss
@@ -5,7 +5,7 @@
   vertical-align: middle;
 }
 :host(:hover) {
-  background-color: var(--kd-color-background-ui-soft);
+  background-color: var(--kd-color-background-ui-soft) !important;
 }
 
 :host([selected]:not([disabled]):not([preventHighlight])),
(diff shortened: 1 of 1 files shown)


maulik-thaker1:
UX Approved by @kaanhsu

src/components/reusable/table/table-row.scss
```diff
@@ -4,6 +4,9 @@
   display: table-row;
   vertical-align: middle;
 }
+:host(:hover) {
+  background-color: var(--kd-color-background-ui-soft) !important;
```
brian-patrick-3, line 8: Can it be done without `!important`? 
I think it might be nice to also add a little `transition` like `background-color 150ms ease-out`.

maulik-thaker1: Nope I tried but so many css overrides inside table. As table has so many :hover properties overriden like active (hover) disabled etc.
sure we can add transition.

maulik-thaker1: adding 200ms. I think for table it's 200ms by peter in rest of the places.

PR #212, opened 2024-08-06
style(table): Change table rollover color as per UX updates

maulik-thaker1:
## Summary

Change table roll-over color to --kd-color-background-inverse-hover as per UX updates (at 06th Aug 2024)


maulik-thaker1:
This is on discussion with UX (@kaanhsu) about whether we gonna provide this optionally or default.

maulik-thaker1:
> This is on discussion with UX (@kaanhsu) about whether we gonna provide this optionally by boolean prop. or default.

After discussion internally with team, we are going to apply this roll over bydefault to table row

PR #226, opened 2024-09-04
fix(Tooltip): on-tooltip-toggle emit event listeners

kumarivijeta:
## Summary

**RCA**
Event listeners were not getting emitted on-tooltip-toggle due to typo of text "open" which was used on updated() method to _emitToggle() based on conditional check.

**Solution**
Update the text "open" to "_open" which solves the emit toggle action.

## Screenshots

**Before Fix**

image

**After Fix**

image


jaideep08 (approved):

srpriyankashetty (approved):
Able to see the Actions populating in the section when hover over the tooltip.!

image

PR #231, opened 2024-09-09
fix: remove unnecessary bounding contexts created by transforms

brian-patrick-3:
## Summary

`transfrom`s create a bounding context which throws off the placement of absolute and fixed position elements. Example: using a Tooltip inside of a Modal. Switching the open states to `transform: none` maintains the transitions/animations while also removing that bounding context.


kumarivijeta (approved):
Looks Good .

PR #234, opened 2024-09-12
fix(Input): accessibility enhancements for Number & Text Input, Time, Date & DateRange Picker

kumarivijeta:
## Summary

Accessibility enhancement for Number & Text Input, Time, Date & DateRange Picker

NA

## Checklist

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -1139,2661 +1139,6 @@
     },
     {
       "kind": "javascript-module",
-      "path": "src/components/global/footer/footer.ts",
-      "declarations": [
-        {
-          "kind": "class",
-          "description": "The global Footer component.",
-          "name": "Footer",
-          "slots": [
-            {
-              "description": "Default slot, for links.",
-              "name": "unnamed"
-            },
-            {
-              "description": "Slot for the logo, will overwrite the default logo.",
-              "name": "logo"
-            },
-            {
-              "description": "Slot for the copyright text.",
-              "name": "copyright"
-            }
-          ],
-          "members": [
-            {
-              "kind": "field",
-              "name": "rootUrl",
-              "type": {
-                "text": "string"
-              },
-              "default": "'/'",
-              "description": "URL for the footer logo link. Should target the application home page.",
-              "attribute": "rootUrl"
-            },
-            {
-              "kind": "method",
-              "name": "handleRootLinkClick",
-              "privacy": "private",
-              "parameters": [
-                {
-                  "name": "e",
-                  "type": {
-                    "text": "Event"
-                  }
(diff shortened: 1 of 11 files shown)


brian-patrick-3 (changes requested):
Use the deepmerge approach whenever using the textStrings object containing one or more keys/strings. This prevents default keys getting dropped when not all keys are provided externally.

brian-patrick-3 (approved):

srpriyankashetty:
Observations
1. The mandatory error message does not say whether user must fill out the Start Date or End date, just says Please fill out this field.
2. Once the mandate error message starts showing, it read out the error after filling every field. Ex: "Month End date spine button 09 Please fill out this field" or "Year End date spine button 2024 Please fill out this field". This is bit confusing when it comes to accessibility.
3. The year field in all the date picker takes 6-digit input or more in Date picker
4. Field validation - If tried to correct the time or any field in date, the whole field is getting erased.

brian-patrick-3:
> Observations
> 
>     1. The mandatory error message does not say whether user must fill out the Start Date or End date, just says Please fill out this field.
> 
>     2. Once the mandate error message starts showing, it read out the error after filling every field. Ex: "Month End date spine button 09 Please fill out this field" or "Year End date spine button 2024 Please fill out this field". This is bit confusing when it comes to accessibility.
> 
>     3. The year field in all the date picker takes 6-digit input or more in Date picker
> 
>     4. Field validation - If tried to correct the time or any field in date, the whole field is getting erased.

1. This is a symptom of a component that is treated as a single input having two input fields within it. To inherit native validation messages, we can only connect to one of the internal inputs at a time, and the messages are not customizable. I don't think we can really do anything about this.
2. Same issue as number 1.
3. These are native HTML inputs, so this behavior is as expected.
4. This might warrant further investigation.

srpriyankashetty:
There is a particular case where it is failing to read the error icon.
Enter the text and erase it, thought the Error icon populates, it reads only the error message.
When clicked out in the plain area and tried to navigate across the fields again, then the error icon is read as Error. Followed by error message.
Based on the discussion, this issue is observed on MAC as well.!

brian-patrick-3:
> There is a particular case where it is failing to read the error icon. Enter the text and erase it, thought the Error icon populates, it reads only the error message. When clicked out in the plain area and tried to navigate across the fields again, then the error icon is read as Error. Followed by error message. Based on the discussion, this issue is observed on MAC as well.!

Might just be a quirk of the screen reader. My only suggestion would be to move the error icon down in front of the actual error message (like Dropdown). Eventually we can do the same for all form input components.

PR #236, opened 2024-09-16
feat: counter for textArea and textInput

tulasigudibanda:
## Summary

Moved char counter below the Text Area.
Add char counter below the Text Input and only show counter when maxLength is provided.

https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/1918161

## Checklist

## Screenshots

image

### DIFF
--- a/src/components/reusable/textArea/textArea.scss
+++ b/src/components/reusable/textArea/textArea.scss
@@ -24,11 +24,14 @@ textarea {
   right: 16px;
 }
 
-.count {
-  @include typography.type-ui-03;
-  position: absolute;
-  right: 22px;
-  bottom: 6px;
-  padding: 2px;
-  background: var(--kd-color-background-inverse);
+.caption-error-count {
+  display: flex;
+  align-items: baseline;
+  justify-content: space-between;
+  .count {
+    @include typography.type-ui-03;
+    margin-top: 13px;
+    margin-left: auto;
+    background: var(--kd-color-background-inverse);
+  }
 }
--- a/src/components/reusable/textArea/textArea.scss
+++ b/src/components/reusable/textArea/textArea.scss
@@ -28,10 +28,9 @@ textarea {
   display: flex;
   align-items: baseline;
   justify-content: space-between;
+  gap: 1rem;
   .count {
     @include typography.type-ui-03;
     margin-top: 13px;
-    margin-left: auto;
-    background: var(--kd-color-background-inverse);
   }
 }
--- a/src/components/reusable/textArea/textArea.scss
+++ b/src/components/reusable/textArea/textArea.scss
@@ -30,7 +30,8 @@ textarea {
   justify-content: space-between;
   gap: 1rem;
   .count {
-    @include typography.type-ui-03;
+    @include typography.type-ui-02;
+    color: var(--kd-color-text-secondary);
     margin-top: 13px;
   }
 }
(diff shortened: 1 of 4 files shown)


src/components/reusable/textArea/textArea.scss
```diff
@@ -24,11 +24,14 @@ textarea {
...
 
-.count {
-  @include typography.type-ui-03;
-  position: absolute;
-  right: 22px;
-  bottom: 6px;
-  padding: 2px;
-  background: var(--kd-color-background-inverse);
+.caption-error-count {
+  display: flex;
```
brian-patrick-3, line 28: Add a `gap: 16px;` or similar, so the caption doesn't run right into the count if long enough.

src/components/reusable/textArea/textArea.scss
```diff
@@ -24,11 +24,14 @@ textarea {
...
+  align-items: baseline;
+  justify-content: space-between;
+  .count {
+    @include typography.type-ui-03;
+    margin-top: 13px;
+    margin-left: auto;
+    background: var(--kd-color-background-inverse);
```
brian-patrick-3, line 35: Seems to be no need for these properties any longer.

src/components/reusable/textInput/textInput.scss
```diff
@@ -89,3 +90,15 @@ input[type='search']::-webkit-search-results-decoration {
     display: block;
   }
 }
+
+.caption-error-count {
```
brian-patrick-3, line 94: Same comments as for textarea.

brian-patrick-3 (changes requested):

src/components/reusable/textArea/textArea.scss
tulasigudibanda: Done

src/components/reusable/textInput/textInput.scss
tulasigudibanda: Done

src/components/reusable/textArea/textArea.scss
tulasigudibanda: Removed  margin-left and background . 
Retained margin-top to prevent the shifting of count div in absence of caption/invalidText.

brian-patrick-3 (approved):

srpriyankashetty (approved):
Verified the counter text and max length and looks fine 
Also verified the following attributes in Text area ans Text input

Verified changing the Label and reflects in the view area
Verified setting the value
Verified setting the placeholder
Verified setting the caption
Verified enabling the required field, the * icon, error icon and error message are displayed
Verified disabling the field and looks fine
Verified setting the Invalid text and it displays all the time
Verified setting the min length and error shows until the min characters are entered.
Verified setting the max length and restricts until those many characters
Verified setting the rows, and the text area height changes based on the rows added
Verified enabling hide label option and looks fine

Text input and with icon attributes
Verified setting the size to small, medium and large
Verified that icon can be placed right or left
Verified setting the pattern and error message shows as "Please match the requested format." if pattern is not matched
**But when more than one character is entered with same patter, the same error message is showing again** 

image

brian-patrick-3:
> Verified setting the pattern and error message shows as "Please match the requested format." if pattern is not matched **But when more than one character is entered with same patter, the same error message is showing again**

I think that's just an issue with your RegEx pattern, `[a-z]*` will work for multiple characters.

PR #238, opened 2024-09-18
feat(progress-bar): add shidoka progress bar status indicator component

wbarbee:
## Summary

Adds progress bar status indicator component -- `<kyn-progress-bar>` (UX currently pending).

[Link here (if applicable).](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/1918195)

## Checklist

## Screenshots

(if any)

### DIFF
--- a/src/components/reusable/progressBar/progressBar.scss
+++ b/src/components/reusable/progressBar/progressBar.scss
@@ -1,0 +1,164 @@
+@use '../../../common/scss/global.scss';
+@use '~@kyndryl-design-system/shidoka-foundation/scss/mixins/typography.scss';
+
+$default-value: var(--kd-color-background-primary);
+$success-value: var(--kd-color-background-success);
+$error-value: var(--kd-color-background-error);
+
+$progress-bar-height: 8px;
+$progress-bar-background: var(--kd-color-background-ui-soft);
+
+.progress-bar {
+    &__upper-container {
+        display: flex;
+        justify-content: space-between;
+        align-items: flex-end;
+        margin-bottom: var(--kd-spacing-8);
+    }
+
+    &__label {
+        margin: 0;
+        color: var(--kd-color-text-secondary);
+        @include typography.type-ui-01;
+        font-weight: 500;
+    }
+
+    &__status-icon {
+        padding-right: 4px;
+        line-height: 1;
+
+        kd-icon {
+            display: block;
+        }
+
+        &.success kd-icon {
+            color: $success-value;
+        }
+
+        &.error kd-icon {
+            color: $error-value;
+        }
+    }
+
+    &__item {
+        width: 100%;
+        height: $progress-bar-height;
+        -webkit-appearance: none;
+        appearance: none;
+        border: none;
+        border-radius: 2px;
+        overflow: hidden;
+        background-color: $progress-bar-background;
+
+        &::-webkit-progress-bar {
+            background-color: $progress-bar-background;
+            border-radius: 2px;
+        }
+
+        &::-webkit-progress-value {
+            background-color: $default-value;
+            border-radius: 0 2px 0 2px;
+            transition: width 0.3s ease-in-out;
+        }
+
+        &::-moz-progress-bar {
+            background-color: $default-value;
+            border-radius: 2px;
+            transition: width 0.3s ease-in-out;
+        }
+
+        &::-ms-fill {
+            background-color: $default-value;
+            border: 0;
+        }
+
+        &:indeterminate {
+            background-image: linear-gradient(to right, $default-value 30%, $progress-bar-background 30%);
+            background-size: 200% 100%;
+            animation: indeterminate 1.5s linear infinite;
+
+            &::-webkit-progress-bar {
+                background-color: transparent;
+            }
+
+            &::-moz-progress-bar {
+                background-image: linear-gradient(to right, $default-value 30%, $progress-bar-background 30%);
+                background-size: 200% 100%;
+                animation: indeterminate 1.5s linear infinite;
+            }
+        }
+    }
+
+    &__success {
+        &::-webkit-progress-value {
+            background-color: $success-value;
+        }
+
+        &::-moz-progress-bar {
+            background-color: $success-value;
+        }
+
+        &::-ms-fill {
+            background-color: $success-value;
+        }
+    }
+
+    &__error {
+        &::-webkit-progress-value {
+            background-color: $error-value;
+        }
+
+        &::-moz-progress-bar {
+            background-color: $error-value;
+        }
+
+        &::-ms-fill {
+            background-color: $error-value;
+        }
+    }
+
+    &__helper-text {
+        margin-top: 12px;
+        color: var(--kd-color-text-primary);
+        @include typography.type-ui-04;
+        font-weight: 300;
+
+        &.error {
+            color: $error-value;
+        }
+    }
+
+    &__speed-slow {
+
+        &::-webkit-progress-value,
+        &::-moz-progress-bar {
+            transition: width 0.5s ease-in-out;
+        }
+    }
+
+    &__speed-normal {
+
+        &::-webkit-progress-value,
+        &::-moz-progress-bar {
+            transition: width 0.3s ease-in-out;
+        }
+    }
+
+    &__speed-fast {
+
+        &::-webkit-progress-value,
+        &::-moz-progress-bar {
+            transition: width 0.1s ease-in-out;
+        }
+    }
+}
+
+@keyframes indeterminate {
+    0% {
+        background-position: 200% 0;
+    }
+
+    100% {
+        background-position: -200% 0;
+    }
+}
--- a/src/components/reusable/progressBar/progressBar.ts
+++ b/src/components/reusable/progressBar/progressBar.ts
@@ -1,0 +1,94 @@
+import { html, LitElement } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+import '@kyndryl-design-system/shidoka-foundation/components/icon';
+import checkmarkIcon from '@carbon/icons/es/checkmark--filled/20';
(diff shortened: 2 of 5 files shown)


wbarbee:
Getting the following error from the rollup.config.js:
Screenshot 2024-09-18 at 6 00 05 PM

Found this github issue:
https://github.com/alfredosalzillo/rollup-plugin-multi-input/issues/61

on this branch, i upgraded to v1.5.0 and changed `import multiInput from 'rollup-plugin-multi-input';` --> `import * as multiInput from 'rollup-plugin-multi-input';` as indicated in the Github Issue.

We have a lot of sass warnings, regardless, but the build now passes.

wbarbee:
there might be a bug in the tooltip that i believe only occurs only in the 'Docs' Storybook view. when you hover over the anchor, the overflow-y for the contained view for that variation of the component expands (i think the tooltip is being appended well below the anchor?). this happens on the Progress Bar Docs page as well as the Tooltip Docs page for me

if you click on the individual variation in the sidebar (full view), the tooltip looks as expected

src/components/reusable/progressBar/progressBar.ts
```diff
@@ -0,0 +1,273 @@
...
+  @property({ type: Number })
+  max = 100;
+
+  /** Sets optional progress bar label. */
+  @property({ type: String })
+  label = '';
+
+  /** Sets optionally displayed information icon and tooltip content. */
+  @property({ type: String })
+  informationalTooltipText = '';
```
brian-patrick-3, line 51: Maybe make Tooltip a slot, rather than built-in as a prop. I want to move all form components to have a tooltip slot eventually.

src/components/reusable/progressBar/progressBar.scss
```diff
@@ -0,0 +1,213 @@
+@use '../../../common/scss/global.scss';
+@use '~@kyndryl-design-system/shidoka-foundation/scss/mixins/typography.scss';
```
brian-patrick-3, line 2: I might suggest importing the form-input.scss common styles, which can be leveraged for label, error, & option text for example for consistency.

brian-patrick-3 (changes requested):
There's a pretty large disparity in the styles between Chrome and Firefox, and in Firefox the indeterminate animation does not work.

One thing I find a bit confusing is how the animation works in the active state. What exactly drives it to increase progress? I imagine with a download/upload for example, the value would have to be updated at some sort of interval based on data transfer in order to advance it.

brian-patrick-3:
> there might be a bug in the tooltip that i believe only occurs in the 'Docs' Storybook view. when you hover over the anchor, the overflow-y for the contained view for that variation of the component expands (i think the tooltip is being appended well below the anchor?). this happens on the Progress Bar Docs page as well as the Tooltip Docs page for me
> 
> if you click on the individual variation in the sidebar (full view), the tooltip looks as expected

Something to do with how Storybook's Docs pages render the stories inline, messes with fixed/absolute positioning.

wbarbee:
> There's a pretty large disparity in the styles between Chrome and Firefox, and in Firefox the indeterminate animation does not work.
> 
> One thing I find a bit confusing is how the animation works in the active state. What exactly drives it to increase progress? I imagine with a download/upload for example, the value would have to be updated at some sort of interval based on data transfer in order to advance it.

apparently firefox doesn't allow pseudo-elements/classes inside of indeterminate state, while chrome and safari do 😤 ... so i'm going to have to figure out whether i can make it work using <progress>, given this limitation.

on your second point, i think i see what you're saying. i've tried to keep this as generic as possible. i was envisioning that the `value` property would be relied on and that the dev using Shidoka would listen for events that would progress that `value` prop.

brian-patrick-3:
> apparently firefox doesn't allow pseudo-elements/classes inside of indeterminate state, while chrome and safari do 😤 ... so i'm going to have to figure out whether i can make it work using , given this limitation.

Unfortunate :(

> i was envisioning that the `value` property would be relied on and that the dev using Shidoka would listen for events that would progress that `value` prop.

OK, that's what I was picturing as well. If the value gets increased, does it animate then? Ex: jump from 0 to 50.

src/components/reusable/progressBar/progressBar.scss
```diff
@@ -0,0 +1,212 @@
...
+        background-color: $color;
+    }
+
+    &::-ms-fill {
+        background-color: $color;
+    }
+}
+
+:host {
+    display: block;
```
brian-patrick-3, line 30: Might want to default to inline-block, I can imagine use cases for having it inline.

src/components/reusable/progressBar/progressBar.scss
```diff
@@ -0,0 +1,212 @@
...
+    font-family: 'Roboto', Arial, Helvetica, sans-serif;
```
brian-patrick-3, line 31: This should be unnecessary, font-family is inherited.

wbarbee:
> > apparently firefox doesn't allow pseudo-elements/classes inside of indeterminate state, while chrome and safari do 😤 ... so i'm going to have to figure out whether i can make it work using , given this limitation.
> 
> Unfortunate :(
> 
> > i was envisioning that the `value` property would be relied on and that the dev using Shidoka would listen for events that would progress that `value` prop.
> 
> OK, that's what I was picturing as well. If the value gets increased, does it animate then? Ex: jump from 0 to 50.

yes, so if the value is increased to 50 (65% in the default example), it will smoothly increase to that value and stop when it's reached.

i was kind of in the same mindset re: the redundancy of the progress spinner in the upper right. but i guess it makes sense to optionally include it. currently, if you hardcode a value like in the default example, the spinner disappears when the desired value is reached. in the case of success or error, that inline loading spinner thing is replaced by the appropriate icon.

src/components/reusable/progressBar/progressBar.ts
```diff
@@ -0,0 +1,273 @@
...
+    const helperText = this.getHelperText();
+
+    this._percentage = Math.round((this._progress / this.max) * 100);
+
+    return html`
+      ${this.renderProgressBarLabel(currentStatus)}
+      ${this.renderProgressBar(currentStatus, currentValue)}
+      ${helperText
+        ? html`<div class=${`progress-bar__helper-text ${currentStatus}`}>
+            <h2>${helperText}</h2>
```
brian-patrick-3, line 96: Don't bake-in any heading tags, since the component won't be aware of the overall page hierarchy.

src/components/reusable/progressBar/progressBar.ts
```diff
@@ -0,0 +1,273 @@
...
+      )}
+      aria-valuemin="0"
+      aria-valuemax=${this.max}
+      aria-label=${this.label}
+    ></progress>`;
+  }
+
+  private renderProgressBarLabel(currentStatus: ProgressStatus) {
+    return html`<div class="progress-bar__upper-container">
+      <h2 class="progress-bar__label">
```
brian-patrick-3, line 131: Should use a `<label>` tag, and associate it with the `<progress>` tag for accessibility.

src/components/reusable/progressBar/progressBar.ts
```diff
@@ -0,0 +1,273 @@
...
+import { customElement, property, state } from 'lit/decorators.js';
+import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+import '@kyndryl-design-system/shidoka-foundation/components/icon';
+import '../loaders';
+import '../tooltip';
+import checkmarkIcon from '@carbon/icons/es/checkmark--filled/16';
+import errorIcon from '@carbon/icons/es/error--filled/16';
+import informationIcon from '@carbon/icons/es/information/16';
```
brian-patrick-3, line 11: I was going to say we should use Shidoka icons here, but I don't think there's an error icon yet.

src/components/reusable/progressBar/progressBar.ts
```diff
@@ -0,0 +1,273 @@
...
+      [`progress-bar__${status}`]: true,
+    });
+  }
+
+  private getHelperText() {
+    if (this.helperText) {
+      return this.helperText;
+    }
+
+    if (this.status === ProgressStatus.ACTIVE) {
```
brian-patrick-3, line 198: Might be an issue if there's no way to hide this active state helper text?

src/components/reusable/progressBar/progressBar.ts
```diff
@@ -0,0 +1,273 @@
...
+      const formattedProgress =
+        this.unit === '%'
+          ? this._progress.toFixed(0)
+          : this._progress.toFixed(1);
+      const formattedMax =
+        this.unit === '%' ? this.max.toFixed(0) : this.max.toFixed(1);
```
brian-patrick-3, line 204: Might be simpler to just use the raw values provided for `value` and `max`.

src/components/reusable/progressBar/progressBar.ts
```diff
@@ -0,0 +1,273 @@
...
+export class ProgressBar extends LitElement {
+  static override styles = ProgressBarStyles;
+
+  /** Sets visibility of optional inline load status spinner. */
+  @property({ type: Boolean })
+  showInlineLoadStatus = true;
+
+  /** Sets progress bar status mode. */
+  @property({ type: String })
+  status: 'indeterminate' | 'active' | 'success' | 'error' = 'indeterminate';
```
brian-patrick-3, line 35: Does 'indeterminate' need to be a state, or can it be derived from absence of `value`?

brian-patrick-3 (commented):
Had some more thoughts after looking at it a bit more closely.

src/components/reusable/progressBar/progressBar.ts
wbarbee: in this case, `_progress` is used internally for the smooth animations, incrementing the bar. while `value` could be just a static, final numeric value.

if it's alright, i'll just strip the formatting and use the raw `this._progress` and `this.max` values

src/components/reusable/progressBar/progressBar.ts
wbarbee: added `showActiveHelperText`

PR #241, opened 2024-09-25
feat(progress-bar): component constructed with nested divs, instead of <progress> element

wbarbee:
## Summary

Adds progress bar status indicator component -- `<kyn-progress-bar>`

[UX Designs](https://www.figma.com/design/AIX4LLzoDHnFCXzQCiPHJk/Vienna?node-id=4308-129&node-type=canvas&t=4PoWDTrfawHFRKsj-0)

UI: progress bar [[link]](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/1918195)

## Notes

- uses nested divs, instead of <progress> HTML element
- this PR also cleans up some Typescript github warnings for unused variables
- this PR also addresses a failed build due to an incompatibility in the rollup.config (see comment below)

## Checklist

## Screenshots

Screenshot 2024-09-25 at 8 19 52 AM
Screenshot 2024-09-25 at 8 19 44 AM
Screenshot 2024-09-25 at 8 19 58 AM
Screenshot 2024-09-25 at 8 19 37 AM
Screenshot 2024-09-25 at 8 19 28 AM


wbarbee:
code review issues from previous <progress> PR addressed here:

- tooltip slotted, instead of built-in prop
- leveraging form-input-scss for helper text, label, error, success styling for consistency
- :host display block converted to inline-block to allow for inline placement
- hardcoded 'Roboto' font-family removed, as it's inherited
- convert heading tags to <span>, since component won't be aware of the overall page hierarchy
- associated <label> to the progress bar component (now nested <div>'s) with `for=""` and target id
- shidoka icons will replace carbon icons when available
- added `showActiveHelperText` property to hide active state helper text
- removed `.toFixed()` decimal formatting, using raw values instead
- indeterminate state is now inferred

src/components/reusable/progressBar/progressBar.scss
```diff
@@ -0,0 +1,168 @@
...
+
+    &__active {
+        border-top-right-radius: 0;
+        border-bottom-right-radius: 0;
+    }
+}
+
+@keyframes indeterminate {
+    0% {
+        left: 0;
```
brian-patrick-3, line 136: This animation is fine as is IMO, but animating a property like left is probably not super smooth compared to transform for example especially on low-end devices.

src/components/reusable/progressBar/progressBar.ts
```diff
@@ -0,0 +1,285 @@
...
+export class ProgressBar extends LitElement {
+  static override styles = ProgressBarStyles;
+
+  /** Sets visibility of optional inline load status spinner. */
+  @property({ type: Boolean })
+  showInlineLoadStatus = true;
+
+  /** Controls whether to show default helper text for active state. */
+  @property({ type: Boolean })
+  showActiveHelperText = true;
```
brian-patrick-3, line 34: Boolean props must default to false, ref: https://lit.dev/docs/components/properties/#boolean-attributes

brian-patrick-3 (changes requested):
Shaping up nicely, had a few more small comments.

src/components/reusable/progressBar/progressBar.ts
wbarbee: ✅

src/components/reusable/progressBar/progressBar.scss
```diff
@@ -0,0 +1,168 @@
...
+    display: inline-block;
+    width: 100%;
+}
+
+::slotted([slot="unnamed"]) {
+    display: inline-block;
+    line-height: 8px;
+    margin-left: 0;
+    margin-top: 1px;
+}
```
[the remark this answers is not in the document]
wbarbee: having some issues breaking through to definitively style just the icon. this seeks to simply vertically align the tooltip anchor to the label text. is it alright to use inline styles in this case on a storybook slotted element? seems like maybe no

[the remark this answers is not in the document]
brian-patrick-3: I think `::slotted` is the right approach, just the `line-height` inheritance is what concerned me. 

Also, you don't actually need to specify the name for the "unnamed" slot. Ex:
`::slotted(kyn-tooltip)` or `slot[name="tooltip"]::slotted(kyn-tooltip)` is a bit cleaner

src/common/scss/form-input.scss
```diff
@@ -33,6 +33,22 @@
   display: flex;
 }
 
+.helper-text {
+  margin: 0;
+
+  span {
+    color: var(--kd-color-text-secondary);
+    font-size: 12px;
```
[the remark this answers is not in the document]
brian-patrick-3: Right I'm just forseeing having to change it later after an accessibility audit. For now, just use the type-ui-03 style as designed.

src/components/reusable/progressBar/progressBar.scss
wbarbee: lastly, `translateX()` is a bit tricky here, as it executes the animation based on the element's width that it's attached to, not the parent's width.

it would take some restructuring, but i think i could make it work. i understand that it leads to better performance.

brian-patrick-3: Right, I had the same thought, just wanted to note that. Don't waste any time on it if it doesn't make sense. It does work fine as is.

wbarbee: All good. These are all good things to consider, thank you

brian-patrick-3 (approved):

srpriyankashetty (approved):
1. Verified the Default, Indeterminate, Simulated success and Error Progres bar
2. Verified the Active helper text, true or false
3. Verified changing the Status, the progress bar color is changing based on the status selected.
4. Verified the progress bar label, tooltip, helper text and looks fine
5. Verified changing the Max value and Units and changes are reflecting in view area.

srpriyankashetty:
In Indeterminate progress bar observed null% of null% showing, is that expected?

image

wbarbee:
> In Indeterminate progress bar observed null% of null% showing, is that expected?
> 
> image

added an additional check to ensure `null% of null%` does not show. thank you!

PR #243, opened 2024-10-02
feat(local-nav)!: local nav redesign

brian-patrick-3:
## Summary

Redesign/rework of the local nav component.

## Figma Link

https://www.figma.com/design/AIX4LLzoDHnFCXzQCiPHJk/Vienna?node-id=4018-504&node-type=canvas&m=dev

## ADO Story Link

https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/1935789

## Notes

- [x] Flyouts changed to expand/collapse
- [x] Support for 4 levels instead of 3
- [x] Add a Divider sub-component with an optional prop for heading/category.
- [x] Add a slot for search
   - Search functionality must be provided at the application level
      - [x] Add a helper function for filtering a nested array of links if possible.
   - [x] Create a Story showing a functional search filter


src/components/global/localNav/localNavLink.ts
```diff
@@ -85,51 +93,39 @@ export class LocalNavLink extends LitElement {
...
-   */
-  @state()
-  menuPosition: any = {};
-
   override render() {
     const classes = {
       'level--1': this._level == 1,
       'level--2': this._level == 2,
       'level--3': this._level == 3,
+      'sub-level': this._level > 1,
```
wbarbee, line 101: comparing to the figma, there isn't a fourth level with the filled in bullet to review in the storybook. assuming that's taken care of, or maybe it was deemed unnecessarily in the nesting?

src/components/global/localNav/localNav.stories.js
```diff
@@ -38,50 +46,234 @@ export const LocalNav = {
...
-          L2 Link 2
-        </kyn-local-nav-link>
-      </kyn-local-nav-link>
 
-      <kyn-local-nav-link href="javascript:void(0)" expanded>
-        <kd-icon slot="icon" .icon=${sampleIcon}></kd-icon>
-        Link 3
-
-        <kyn-local-nav-link slot="links" href="javascript:void(0)">
+        <kyn-local-nav-link href="javascript:void(0)">
```
wbarbee, line 62: i assume that these <kyn-local-nav-links> are repeated for readability and because this is the storybook file? that would make sense. otherwise, using something like the `renderLink` function to DRY this up would be good.

src/components/global/localNav/localNav.ts
```diff
@@ -157,6 +168,10 @@ export class LocalNav extends LitElement {
       link._navExpanded = this._expanded || this.pinned;
       link._navExpandedMobile = this._mobileExpanded;
     });
+
+    this._dividers.forEach((divider: any) => {
+      divider._navExpanded = this._expanded || this.pinned;
```
wbarbee, line 173: could add an `on-expand` or `on-pin` action to expand on `on-toggle`

src/components/global/localNav/localNav.scss
```diff
@@ -73,6 +72,22 @@ nav {
   }
 }
 
+.search {
```
wbarbee, line 75: the storybook variations differ from the figma in spacing of the search bar and the padding/margin between the top of the nav and the first link. in the actual implementation it could be different, just noting that here

wbarbee (commented):
this looks good to me. just a couple of notes or suggestions on things that were likely intentional and logical on your part.

src/components/global/localNav/localNav.scss
wbarbee: this could also be awkwardness in the design that i'm feeling ... maybe the contrast of the size of the magnifying glass search icon, general scale of the search input, and the icons/nodes below it and how they're aligned.

src/components/global/localNav/localNavLink.ts
brian-patrick-3: Cleaned this up a bit. The updates kinda made nesting recursive/infinite, although the top level still has some unique considerations with the collapsed nav state.

src/components/global/localNav/localNav.stories.js
brian-patrick-3: Mainly so the example code is easy to see in storybook. If we start abstracting it, it makes it hard to find. You're right though, most likely the application will render links in a loop rather than hard-code each tag.

src/components/global/localNav/localNav.ts
brian-patrick-3: `on-toggle` just emits the new pinned state value. An event for the expand/collapse on mouse in/out probably doesn't have a use case that I can think of at least.

src/components/global/localNav/localNav.scss
brian-patrick-3: Adjusted the search input padding slightly, seems like Figma had an extra 2px around it.

wbarbee (approved):

wbarbee:
lgtm

tulasigudibanda (approved):

srpriyankashetty:
Hi Brian,
1. I see that the alignment is missing when the options are without and with next level
2. The Figma shows solid O icon for the one in expanded state and collapsed one as O, is this expected?
3. In Figma, there are other icons included but storybook shows only O. Do we have any code level option to modify these as per business requirement? 
4. Is there an option to change the Divider with heading in the code level? Since I do not see any attribute in Controls.

image
image

brian-patrick-3:
> Hi Brian,
>
>     1. I see that the alignment is missing when the options are without and with next level
>     2. The Figma shows solid O icon for the one in expanded state and collapsed one as O, is this expected?
>     3. In Figma, there are other icons included but storybook shows only O. Do we have any code level option to modify these as per business requirement?
>     4. Is there an option to change the Divider with heading in the code level? Since I do not see any attribute in Controls.

1. I believe this is the correct alignment, +/- aligns with icon if no sub-menu items are present.
2. The circle icon is just a placeholder. Icons are only required for top level menu items.
3. Same answer as 2.
4. Since the Divider is a sub-component, I did not set up Controls for it, and just added one hard-coded example Story.

PR #256, opened 2024-10-14
fix(stepper): add support for link url values on stepper items

brian-patrick-3:
## Summary

A bug was raised that stepper item title links could not be controlled by the application. This PR accomplishes the following:

1. Added a component prop so a link href can be passed to stepper items/children.
2. Step Title will render as plain text if no link is provided.
3. Link href value added to click event details.
4. Child step items will now always be accessible/expandable even when the parent step is not in active state.

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -680,7 +680,7 @@
                 "text": "string"
               },
               "default": "''",
-              "description": "DEPRECATED. Use `label` instead.\nButton assistive text, title + aria-label.",
+              "description": "DEPRECATED. Use `label` instead.\r\nButton assistive text, title + aria-label.",
               "attribute": "assistiveText"
             },
             {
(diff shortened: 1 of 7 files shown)


tulasigudibanda (approved):

marcosebastianlopez (approved):
This looks good, thank you for doing this.

srpriyankashetty:
In case of Horizontal stepper, the boarder padding must be taken care when clicked for all the 4-link text, rest other pages looks fine.

image

brian-patrick-3:
> In case of Horizontal stepper, the boarder padding must be taken care when clicked for all the 4-link text. 
> And the alignment between step number and the status is different for every status. Rest other pages look fine.
1. Added 1px padding to accommodate link focus outline.
2. first/last steps have left/right alignments, the rest are centered, as per the design.

srpriyankashetty (commented):
Verified the link text in Horizontal stepper
Verified the link text in Vertical stepper
Verified the link text in Nested stepper
Verified that link is not provided for the step that are not completed
Verified that href is passed to stepper items/children.
Verified the Step Title renders as plain text if there is no link provided.
Verified the Link href value is added to click event details.
Verified that Child step items are now accessible/expandable even when the parent step is not in active state.
Verified the Stepper type Procured and status format

PR #273, opened 2024-11-07
feat(skeleton): vital card

kumarivijeta:
## Summary

Addad a control to show multiple skeleton and Skeleton added for Vital Card

Link here (if applicable).

## Figma Link

Link here (if applicable).

## Checklist

## Testing Instructions

## Screenshots

(if any)

### DIFF
--- a/src/components/reusable/loaders/skeleton.ts
+++ b/src/components/reusable/loaders/skeleton.ts
@@ -13,8 +13,16 @@ export class Skeleton extends LitElement {
   @property({ type: Boolean })
   inline = false;
 
+  /** Number of skeleton lines to show. */
+  @property({ type: Number })
+  lines = 1;
+
   override render() {
-    return html` <div class="skeleton"></div> `;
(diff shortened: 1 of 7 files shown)


src/components/reusable/loaders/skeleton.ts
```diff
@@ -13,8 +13,16 @@ export class Skeleton extends LitElement {
   @property({ type: Boolean })
   inline = false;
 
+  /** Number of skeleton lines to show. */
+  @property({ type: Number })
+  lines = 1;
```
brian-patrick-3, line 18: I like this idea.

brian-patrick-3 (changes requested):
A few thoughts:

1. We don't need to make skeletons for form input components, like Datepicker.
2. I think Skeleton versions of a component should be completely separate components/files. Ex: don't put skeleton inside tags, create a separate tags skeleton file.
3. For Tag, maybe build it for the Tag Group instead of an individual tag.

PR #276, opened 2024-11-20
fix: disabled input hover & icon color

brian-patrick-3:
## Summary

Fixed the disabled form input styles so the border color doesn't change on hover, and the text input icon uses the disabled text color.


tulasigudibanda (approved):

wbarbee (approved):

srpriyankashetty (commented):
Verified that icon is grayed out on disable in Text Area
Verified that on disabling boarder is not changed on hover
Verified that whole text area is grayed out when disabled

PR #287, opened 2024-12-03
fix(code view): dark mode token updates

wbarbee:
## Summary

Coalesce dark and light syntax themes into a united theme that relies on new tokens with the following rules:

1) default to the global <meta> `color-scheme` attribute setting for system-based dark or light mode
2) maintain a `darkTheme` property that the dev can set to override the theme when desired

[Code View ADO Task](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2003592)

## Figma Link

[Code View Updated Figma](https://www.figma.com/design/CQuDZEeLiuGiALvCWjAKlu/branch/qMpff4GuFUEcsMUkvacS3U/Applications---Component-Library?node-id=16748-7660&node-type=canvas&m=dev)

## To Do

- [x] `darkTheme` override hierarchy
- [x] check with UX on light background hex value for code view (it's not purely white)
- [x] delete old syntax files when done

## Testing Instructions
- currently, UX needs to update the token hex value for the light theme background:
`--kd-color-code-view-background | light-dark(#fafafc, #222527)`

this will be updated as soon as available

## Checklist

## Screenshots
Screenshot 2024-12-09 at 10 29 40 AM
Screenshot 2024-12-09 at 10 29 46 AM


src/components/reusable/blockCodeView/blockCodeView.ts
```diff
@@ -47,15 +46,11 @@ const LANGUAGE_SPECIFIC_TOKENS: Record<string, string[]> = {
...
-    BlockCodeViewStyles,
-    ShidokaLightTheme,
-    ShidokaDarkTheme,
-  ];
+  static override styles = [BlockCodeViewStyles, ShidokaSyntaxTheme];
 
   /** Sets background and text theming. */
   @property({ type: String })
-  darkTheme: 'light' | 'dark' = 'dark';
+  darkTheme: 'light' | 'dark' | 'default' = 'default';
```
brian-patrick-3, line 53: I notice when changing this, it's changing the theme of the entire page. Looks like it's targeting the meta tag? We don't want that to happen here. This should really be an override which only applies to the code snippet itself. That can be achieved by using CSS `color-scheme: <value>` on the container element.

brian-patrick-3 (changes requested):

src/components/reusable/blockCodeView/blockCodeView.ts
wbarbee: i refactored both inline and block components and how they're handling this separation and prioritization of theming coming from the global `color-scheme` vs. the component level `darkTheme` property. this should be fixed on the deploy preview now.

wbarbee:
there are a few questions for UX outstanding:

- [ ] 1. **looks like the [copy and expand buttons](https://www.figma.com/design/CQuDZEeLiuGiALvCWjAKlu/branch/qMpff4GuFUEcsMUkvacS3U/Applications---Component-Library?node-id=16748-11543&node-type=instance&m=dev) in the figma file are no longer buttons but are instead links. i've converted the kd-buttons to divs so there won't be any animation, so it will simply have hover and disabled states to style**
Screenshot 2024-12-11 at 12 22 35 PM
Screenshot 2024-12-11 at 12 22 37 PM
Screenshot 2024-12-11 at 12 23 12 PM
Screenshot 2024-12-11 at 12 23 10 PM

- [ ] 2. **still some accessibility errors (i know this is still being designed):**
Screenshot 2024-12-11 at 12 22 12 PM
Screenshot 2024-12-11 at 12 24 36 PM

 

- [ ] 3. `background` **color hex value for** `--kd-color-code-view-background` **still seems a bit light (discussed this with Robert)**
Screenshot 2024-12-11 at 12 22 30 PM


cc: @kaanhsu

brian-patrick-3:
1. I still see the copy link as a Tertiary button in Figma?
2. Hmm we didn't actually change the syntax colors at all, did we? I assumed those were already accessible?

src/components/reusable/blockCodeView/blockCodeView.ts
brian-patrick-3: It seems like it's scoped correctly now, however the "default" theme is appearing to me as light even though I have auto set overall and my system preference is dark.

wbarbee: i'm not sure all of this prefers logic is necessary, maybe there's already an easier way to achieve this built into the codebase, but let me know your thoughts on:

https://github.com/kyndryl-design-system/shidoka-applications/blob/9529158a46665ad4a5b3707606059088f12d8a58/src/components/reusable/blockCodeView/blockCodeView.ts#L158-L188

brian-patrick-3: I don't think so. By default, you shouldn't have to do anything and it will use the application-level theme. All you should need to do to override that is set CSS `color-scheme: dark/light` on the container.

brian-patrick-3 (approved):
LGTM, pending copy button design.

srpriyankashetty (approved):
Verified and looks fine in both Dark and light mode, 
Please let me know if we are good with Copy button

wbarbee:
i've talked to Robert about these accessibility errors. he's gone through them, updated a few tokens and created some new tokens that Marco might need to push. i still don't think that will fix all of the accessibility errors.

i don't know if these are actually false positives?:
Screenshot 2025-01-08 at 2 30 28 PM
Screenshot 2025-01-08 at 2 30 21 PM

the prism tokens are being applied and read correctly and i've gone through to ensure that they're all being styled as intended, based on [this sharepoint that Robert sent](https://kyndryl-my.sharepoint.com/:x:/r/personal/ka_an_hsu_kyndryl_com/_layouts/15/Doc.aspx?sourcedoc=%7B4A7D82CA-7FF4-434C-BBDD-F46EB639520A%7D&file=Code%20View%20%20Colors.xlsx&action=default&mobileredirect=true).


here is [the css in question](https://github.com/kyndryl-design-system/shidoka-applications/blob/293b2f8f1c3a2e42b6ea6283aef4350e4144b06f/src/common/scss/shidoka-syntax-theme.scss).

@brian-patrick-3, do you think the best way to proceed would be to push those updated tokens and merge this PR, with another accessibility pass through when we're closer to finalizing this 2.0 thing? or just leave it open for now?

cc: @kaanhsu @srpriyankashetty

Issue #292, opened 2024-12-11
Radio button component – Button and label are not grayed out when disabled 


srpriyankashetty:
**Before submitting a bug:**

- [yes, label has issue] Is this reproducible in the latest version of Shidoka? 
- [ yes] Is this reproducible in the [Shidoka Storybook](https://kyndryl-design-system.github.io/shidoka-applications/)?
- [ ] Is this reproducible in a [Sandbox](https://codesandbox.io)?
- If it's not reproducible in any of the above scenarios, please do not open a bug report as it is an application-level issue.

**Describe the bug**

On disabling the Radio button is not greyed out in production, in latest version the label is not greyed out.

**To Reproduce**

1. Go to 'Radio button'
2. Click on 'Disable action'
3. See error

**Expected behavior**

The radio button and the label must be greyed out

**Screenshots**

![image](https://github.com/user-attachments/assets/020dbb34-c9fa-42c3-b79d-7d812e3bb9ad)

**Environment**

**Additional context**

PR #295, opened 2024-12-11
fix(radio button): disable radio button

tulasigudibanda:

https://github.com/kyndryl-design-system/shidoka-applications/issues/292

### DIFF
--- a/src/components/reusable/radioButton/radioButtonGroup.ts
+++ b/src/components/reusable/radioButton/radioButtonGroup.ts
@@ -101,9 +101,12 @@ export class RadioButtonGroup extends FormMixin(LitElement) {
   }
 
   private _updateChildren() {
-    this.radioButtons.forEach((radioButton) => {
-      radioButton.disabled = this.disabled;
-      radioButton.checked = this.value.includes(radioButton.value);
+    this.radioButtons.forEach((radio) => {
+      radio.disabled = this.disabled;
+      radio.checked = radio.value === this.value;
+      radio.name = this.name;
+      radio.required = this.required;
+      radio.invalid = this._isInvalid;
     });
   }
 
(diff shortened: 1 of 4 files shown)


brian-patrick-3 (changes requested):
I'm still seeing radio buttons in the group appearing as enabled on load when the disabled flag is true.

I think the actual fix will be to use `slotchange` event to sync the properties of the group with the child radio buttons. Checkbox group was set up that way, I think radio button never got that treatment.

tulasigudibanda:
> I'm still seeing radio buttons in the group appearing as enabled on load when the disabled flag is true.
> 
> I think the actual fix will be to use `slotchange` event to sync the properties of the group with the child radio buttons. Checkbox group was set up that way, I think radio button never got that treatment.

Pushed the slotchange fix

src/components/reusable/radioButton/radioButtonGroup.ts
```diff
@@ -90,12 +90,24 @@ export class RadioButtonGroup extends FormMixin(LitElement) {
...
-          <slot></slot>
+          <slot @slotchange=${this._handleSlotChange}></slot>
         </div>
       </fieldset>
     `;
   }
 
+  private _handleSlotChange() {
+    this._updateChildren();
+    this.requestUpdate();
```
brian-patrick-3, line 101: Probably don't need to manually request update here. That's usually only necessary if you are using a `query` decorator somewhere in your template, because queries are updated post-render.

src/components/reusable/radioButton/radioButtonGroup.ts
```diff
@@ -90,12 +90,24 @@ export class RadioButtonGroup extends FormMixin(LitElement) {
...
+  }
+
+  private _updateChildren() {
```
brian-patrick-3, line 104: You've got the right idea now, but need to take it a bit further. All of those props within the `updated()` function that cause updates to child radio buttons should also probably go in here. Then you can just call this function in both places.

src/components/reusable/radioButton/radioButtonGroup.ts
tulasigudibanda: Removed

src/components/reusable/radioButton/radioButtonGroup.ts
brian-patrick-3: Right you'd have to call `updateChildren` in both `slotchange` and `updated()`

tulasigudibanda: Pushed the changes

brian-patrick-3 (approved):

srpriyankashetty (approved):
Verified and looks fine

image

PR #330, opened 2025-01-17
feat(carousel): color token updates

tulasigudibanda:
Color token updates for Carousel

[TASK 2003562](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2003562)

## Figma Link

https://www.figma.com/design/CQuDZEeLiuGiALvCWjAKlu/branch/qMpff4GuFUEcsMUkvacS3U/Applications---Component-Library?node-id=16511-14985&p=f&t=ubcUMXhS0pJqnOnW-0


src/common/scss/utility/swiper-shidoka.scss
```diff
@@ -90,28 +90,27 @@
...
+        background-color: var(--kd-color-background-ui-hollow-hover);
       }
 
       &:focus {
-        outline-color: var(--kd-color-border-focus);
+        color: var(--kd-color-text-level-primary);
+        outline-color: var(--kd-color-border-variants-focus);
       }
 
       &-active {
```
brian-patrick-3, line 111: The background color token on the active tab seems wrong. Think it should be `Background/Container/Default`

brian-patrick-3 (changes requested):
There is a problem with the arrows. They do nor change color with the theme, because they are technically background image. `.swiper-button-prev:before` and `.swiper-button-next:before`. Will need to figure out something to address that, though I'm not sure what since we can't target the current mode in CSS.

src/common/scss/utility/swiper-shidoka.scss
tulasigudibanda: @brian-patrick-3  , tab styles in carousel are different from tab component styles. Do you want me to match both and let UX know to update the tokens?

tulasigudibanda:
> There is a problem with the arrows. They do not change color with the theme, because they are technically background images. `.swiper-button-prev:before` and `.swiper-button-next:before`. Will need to figure out something to address that, though I'm not sure what since we can't target the current mode in CSS.


Can we use  background: var(--kd-color-text-label-light-label);  on  .swiper-button-prev, .swiper-button-next   . Arrow is still visible in both modes as the background token is not too dark in dark mode.

src/common/scss/utility/swiper-shidoka.scss
brian-patrick-3: > @brian-patrick-3  , I will correct it. But tab styles in carousel are different from tab component styles. Do you want me to match both and let UX know to update the tokens?

Yeah the colors should match the Tabs line variant.

brian-patrick-3 (approved):

wbarbee (approved):

srpriyankashetty (approved):
Verified and looks fine

PR #334, opened 2025-01-22
fix(overflow-menu): correct fixed horizontal position

brian-patrick-3:
## Summary

Fixed position overflow menu horizontal position was incorrect when inside a scrollable container. Adjusting positioning logic to calculate it correctly when opened.

## Testing Instructions

- Go to Data Table > Sticky Header story
- Shrink the screen so a horizontal scrollbar appears on the table
- Scroll the table right
- Open a row overflow menu, it should be positioned correctly with the anchor button

### DIFF
--- a/src/components/reusable/overflowMenu/overflowMenu.scss
+++ b/src/components/reusable/overflowMenu/overflowMenu.scss
@@ -74,10 +74,5 @@ button {
     &.upwards {
       transform-origin: bottom right;
     }
-
-    &.fixed {
-      right: initial;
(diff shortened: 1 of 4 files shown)


tulasigudibanda (approved):

wbarbee (approved):
👍

srpriyankashetty (approved):
Verified the Action menu or overflow menu in sticky header table
Verified the table scroll right when window is smaller and looks fine
Verified the Anchor button alignment

PR #340, opened 2025-01-27
fix: wb component audit 1

wbarbee:
### Spreadsheet Items Addressed
- [Checkbox group -- Bug 2089092](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2089092): in filterable mode, the ghost button was overlapping with the border of the small input -- converted 'X' close buttons to ghost, `kind="tertiary"` so that border only shows during focus
- [Datepicker/ Date Range Picker/ Timepicker -- Bug 2089092](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2089092): same ghost button overlap issue with same remedy
- All Form Inputs: color contrast accessibility error on disabled `.caption` element -- corrected token being used to match updated Figma
- Global Filter: `outlineOnly` button contrast error with icon and button text -- corrected tokens being used for both to match Figma
- Checkbox Group/Subgroup: when `invalidText` prop changed, the component was being re-rendered and reset back to the default Storybook props -- updated logic so that things are selecting, unselecting, and highlighting as expected
- Datepicker/ Date Range Picker/ Timepicker: error icon aria-label was being incorrectly used on a span wrapper element, so removed to resolve accessibility error
- Textarea:  accessibility warning re: textArea wrapper element having obscured background color -- **this error has been present since prior to 2.0 breaking changes. we'll return to this at a later date**

### Additional Items
- move `.flatpickr-input` (class assigned by 3rd-party library) styles to form-input.scss, style with rest of form inputs
- adjust vertical alignment of datepicker and date range picker current month, year, back and forward arrows

### To Check with UX
- Side Drawer: secondary button contrast issues in light mode
- Toggle button: all, disabled state, color contrast violation with label and background
- Date/date range: light mode only, all variants, color contrast error on `.cur-month`, `.cur-year` elements
- Code view: color contrast on subtitle, update color token value to text primary


tulasigudibanda (approved):

src/common/scss/form-input.scss
```diff
@@ -16,10 +16,6 @@
   &.sr-only {
     @include visibility.sr-only;
   }
-
-  [disabled] & {
-    color: var(--kd-color-text-link-level-disabled);
```
brian-patrick-3, line 21: We're removing the label disabled color?

src/common/scss/form-input.scss
```diff
@@ -86,7 +83,8 @@ textarea {
   &[disabled] {
     background: var(--kd-color-background-ui-hollow-default);
     border-color: var(--kd-color-border-ui-disabled);
-    color: var(--kd-color-text-link-level-disabled);
+    color: var(--kd-color-text-level-disabled);
```
brian-patrick-3, line 86: I think we also need to add `aria-disabled="true"` attributes in the HTML. The contrast checker will ignore elements with this. I'm pretty sure the label doesn't need it because it's associated with the disabled form element.

src/components/reusable/datePicker/datepicker.ts
```diff
@@ -280,11 +281,8 @@ export class DatePicker extends FormMixin(LitElement) {
         @mousedown=${this.preventFlatpickrOpen}
         @click=${this.preventFlatpickrOpen}
       >
-        <span
-          class="error-icon"
-          aria-label=${`${this.errorAriaLabel}` || 'Error message icon'}
-          >${unsafeSVG(errorIcon)}</span
-        >${this.invalidText ||
+        <span class="error-icon">${unsafeSVG(errorIcon)}</span>${this
```
brian-patrick-3, line 284: Hmm we specifically added the `aria-label="Error"` to this icon all of the other inputs to help indicate it's an error message. Maybe it was missing the `role`?

src/components/reusable/globalFilter/globalFilter.sample.ts
```diff
@@ -24,6 +24,7 @@ export class SampleFilterComponent extends LitElement {
   static override styles = css`
     .filter-text {
       display: none;
+      color: var(--kd-color-text-button-light-primary);
```
brian-patrick-3, line 27: This seems to be overriding the button text color? I think a better solution might be to choose a button kind with proper contrast on the filter bar. Or is the tertiary outline button just using the wrong text color token here?

src/components/reusable/globalFilter/globalFilter.sample.ts
```diff
@@ -34,6 +35,7 @@ export class SampleFilterComponent extends LitElement {
 
     span[slot='icon'] {
       display: flex;
+      color: var(--kd-color-icon-primary);
```
brian-patrick-3, line 38: This should really just inherit/match the text color. Not sure why the Figma is like that.

src/components/reusable/textArea/textArea.scss
```diff
@@ -17,4 +17,5 @@ textarea {
   border-radius: 4px;
   margin: 0;
   padding: 16px;
+  background: transparent;
```
brian-patrick-3, line 20: Now the textarea BG doesn't match the other inputs.

brian-patrick-3 (changes requested):

src/common/scss/form-input.scss
wbarbee: this reflects how it appears in the figma file. the disabled labels on these form inputs are not grayed out

Screenshot 2025-01-29 at 10 00 57 AM

brian-patrick-3: Maybe a Figma bug? Pretty sure we want to keep the label disabled color. The [Label page](https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=4311-224340&m=dev) shows disabled styles.

wbarbee: ok, i'll revert that

src/components/reusable/datePicker/datepicker.ts
wbarbee: role works. returning these aria-labels and adding `role="button"`

src/components/reusable/globalFilter/globalFilter.sample.ts
brian-patrick-3: Actually yes I think this will be solved by fixing the Button / Outline issue I logged in the excel

src/common/scss/form-input.scss
wbarbee: the accessibility issues seem to be arising from the contrast of the `.caption` elements. i can add `aria-disabled=${this.disabled}` to the caption itself? simply adding it to the input won't resolve that

brian-patrick-3: Yeah common with disabled colors to not pass contrast, hence needing the aria tag.  You can add it to just the caption, or to the entire wrapper/component and it will cover everything inside.

src/components/reusable/globalFilter/globalFilter.sample.ts
wbarbee: removing these color vals

src/components/reusable/textArea/textArea.scss
wbarbee: Screenshot 2025-01-29 at 10 38 59 AM

this accessibility *warning* is present on `main`, so i'm just going to leave this as is for now, if that's alright

brian-patrick-3: Yeah that's fine. I can't seem to reproduce that though

wbarbee: looks like it's only on Chrome for me, not Firefox or Edge

brian-patrick-3 (approved):

srpriyankashetty:
Hi @wbarbee, i don't see any boarder on hover, the boarder shows only when clicked. Is it the expected behavior?

srpriyankashetty:
Checkbox group -- in filterable mode, the ghost button was overlapping with the border of the small input -- converted 'X' close buttons to ghost, kind="tertiary" so that border only shows during focus  - **works fine**
[Datepicker/ Date Range Picker/ Timepicker -- Bug 2089092 same ghost button overlap issue with same remedy
All Form Inputs: color contrast accessibility error on disabled .caption element -- corrected token being used to match updated Figma - **works fine**
Checkbox Group/Subgroup: when invalidText prop changed, the component was being re-rendered and reset back to the default Storybook props -- updated logic so that things are selecting, unselecting, and highlighting as expected - **works fine**

wbarbee:
hi @srpriyankashetty 

1. Close X button - I don't see any boarder on hover, the boarder shows only when clicked. Is it the expected behavior?
**This will be fixed when we merge in Brian's buttons PR for tertiary button styling, but this is all coming from the button component**

2. Date time picker - kyn-date-picker#date-time-picker-error-message - on disabling date picker without caption - Accessibility issue
**i couldn't reproduce this one, maybe we can screenshare**

3. Text Area - Invalid text is not showing in Red color
**in dark mode, the red for the text is very faint, but the icon wasn't picking up the styles -- this should be fixed now**

4. On enabling Required - 2 accessibility issues are showing
**the aria role warning for the `<abbr>` element has been fixed. the other warning was present prior to these 2.0 changes, so we're going to come back to it later**

srpriyankashetty (approved):
Verified and looks good

PR #341, opened 2025-01-29
fix(Dropdown): disable state fix

kumarivijeta:
## Summary

Dropdown Disable state : Placeholder, caption and invalid text are not grayed out

## ADO Issue Link

[BUG 2089084](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2089084)

## Figma Link

NA

## Checklist

## Screenshots

image

### DIFF
--- a/src/components/reusable/dropdown/dropdown.scss
+++ b/src/components/reusable/dropdown/dropdown.scss
@@ -75,11 +75,16 @@
     background-color: transparent;
 
     &:hover {
-
       [disabled] & {
         border-color: transparent;
       }
     }
+  }
+  &[disabled] {
+    background: var(--kd-color-background-ui-hollow-default);
+    border-color: var(--kd-color-border-ui-disabled);
+    color: var(--kd-color-text-link-level-disabled);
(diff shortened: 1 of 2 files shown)


jaideep08 (approved):

brian-patrick-3:
I think this could actually be fixed in a simpler way. IT seems like this bug was caused by the error and captions divs being moved outside of the `<div class="dropdown">`. You can see if you compare to main branch. If we do it that way, it should re-inherit the shared form-input styles.

brian-patrick-3 (changes requested):
I think this could actually be fixed in a simpler way. It seems like this bug was caused by the error and captions divs being moved outside of the `<div class="dropdown">`. You can see if you compare to main branch. If we do it that way, it should re-inherit the shared form-input styles.

wbarbee:
this was my error in separating the conditional error logic. i'll return it to how it was today

wbarbee:
closing this, as this will be addressed in #342

PR #345, opened 2025-01-31
feat: accessibility fixes for footer, breadcrumbs and widget 

tulasigudibanda:
## Summary

Accessibility fixes for footer, breadcrumbs and widget components.
Spreadsheet : line 14 and line 26
UI fix for widget : Added aria-labelledby to announce the "disabled" state

## Figma Link

https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=10033-1598&p=f&m=dev


src/components/global/footer/footer.ts
```diff
@@ -37,6 +37,7 @@ export class Footer extends LitElement {
             href="${this.rootUrl}"
             class="logo-link"
             @click="${(e: Event) => this.handleRootLinkClick(e)}"
+            aria-label="Go to Homepage"
```
brian-patrick-3, line 40: I think this will need to be customizable, for localization purposes.

src/components/reusable/widget/widget.ts
```diff
@@ -49,7 +49,10 @@ export class Widget extends LitElement {
         <div class="widget-header">
           <slot name="draghandle"></slot>
 
-          <div class="title-desc">
+          <div class="title-desc" aria-labelledby="title-desc-label">
+            <span id="title-desc-label" class="visually-hidden">
+              ${this.disabled ? 'This widget is disabled' : ''}
```
brian-patrick-3, line 54: I don't think we need to add hidden text to spell out that the widget is disabled. Simply adding an `aria-disabled` attribute on the wrapper element should suffice.

brian-patrick-3 (changes requested):

src/components/reusable/widget/widget.ts
tulasigudibanda: aria-disabled did not do anything since it is a div element, unless role is added as below :
<div role="button" aria-disabled="true" ...
Hence I had to use aria-labelledby which screen reader announces it as disabled.

src/components/global/footer/footer.ts
tulasigudibanda: Right Brian . Do you suggest creating a prop?

brian-patrick-3: Yes

src/components/reusable/widget/widget.ts
brian-patrick-3: But that wasn't the issue, the issue was color contrast. Maybe see if there's a role here that sort of fits: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled#associated_roles

tulasigudibanda: Added role="group" . With this screen reader announces the interactive elements in side the disabled widget as dimmed.

src/components/global/footer/footer.ts
tulasigudibanda: Added logoAriaLabel prop

brian-patrick-3: That works. Remember to run `analyze` also to update the docs

src/components/global/footer/footer.ts
```diff
@@ -4,6 +4,7 @@ import { classMap } from 'lit/directives/class-map.js';
 import { unsafeHTML } from 'lit-html/directives/unsafe-html.js';
 import logo from '@kyndryl-design-system/shidoka-foundation/assets/svg/kyndryl-logo.svg';
 import FooterScss from './footer.scss';
+import { ifDefined } from 'lit/directives/if-defined.js';
```
brian-patrick-3, line 7: Is this unused?

src/components/reusable/widget/widget.scss
```diff
@@ -69,3 +70,7 @@
   overflow: auto;
   flex-grow: 1;
 }
+
+.visually-hidden {
+  @include mixins.visually-hidden;
```
brian-patrick-3, line 75: Is this unused?

brian-patrick-3 (approved):

srpriyankashetty:
Hi @tulasigudibanda,
Footer - Observed that screen reader is not reading the logo name
image

PR #366, opened 2025-02-10
feat(modal): ai specific variants

tulasigudibanda:
Added ai specific variant for modal

fixes [AB#2102187](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/2102187)

## Figma Link

[FIGMA](https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=25930-69344&m=dev)

##To Do

Styles for Cancel button - 'X' on Modal ?


src/components/reusable/modal/modal.scss
```diff
@@ -80,32 +86,32 @@ form {
...
-  &:active {
-    border-radius: 2px;
-    border: 1px solid var(--kd-color-border-ui-pressed);
-    background: var(--kd-color-background-ui-hollow-hover);
-  }
-}
+// button {
+//   appearance: none;
+//   background: none;
+//   border: 1px solid transparent;
```
wbarbee, line 92: we can probably remove these, as any new button styles will come from the component that i'm working on.

the ghost variant for the X close button for this AI modal will probably remain the same as the rest of the modals, but confirming that with Marco

src/components/reusable/modal/modal.scss
```diff
@@ -34,6 +34,12 @@ dialog {
   &.size--lg {
     width: 711px;
   }
+
+  &.modal--ai {
+    // @include elevation.shadow('3', true);
+    box-shadow: 0px 0px 24px var(--kd-color-border-dropshadow-ai);
```
wbarbee, line 40: will this use [Brian's new elevation](https://github.com/kyndryl-design-system/shidoka-foundation/pull/98)?

wbarbee (commented):
just a couple of questions, but otherwise looks good

src/components/reusable/modal/modal.scss
tulasigudibanda: Sure William. Looked like the button styles here are deprecated. So I commented out to confirm once. I will remove.

src/components/reusable/modal/modal.scss
tulasigudibanda: Yes. But for the build to pass now to generate Deploy preview , I used local style since elevation PR is not merged yet.

wbarbee (approved):

tulasigudibanda:
> Might need to check with Marco, but I think he mentioned the modal is also going to need a "footer" slot, which is where the chat input will be placed. I think it is meant to replace the buttons.

Sure Brian. I will check

brian-patrick-3:
One other thing to note: there is also a new modal size that needs to be added, large

tulasigudibanda:
> One other thing to note: there is also a new modal size that needs to be added, large

We already have md and lg options for size . Is the new one XL ?

tulasigudibanda:
> > We already have md and lg options for size . Is the new one XL ?
> 
> Oh right... No just add small, and use the values I shared over Teams. Should be fine.

Does small have to different from auto?

brian-patrick-3:
> Does small have to different from auto?

Yes

src/components/reusable/modal/modal.stories.js
```diff
@@ -178,6 +191,135 @@ export const WithForm = {
   },
 };
 
+export const AIConnected = {
```
brian-patrick-3, line 194: This looks more like the full AI Chat Pattern. I think we should set this story aside to use for the pattern later. In the modal stories, you can just use the `aiConnected` prop control to cover this.

src/components/reusable/modal/modal.scss
```diff
@@ -154,6 +136,10 @@ h1 {
   }
 }
 
+::slotted([slot='footer']) {
+  margin-top: 32px;
+}
```
brian-patrick-3, line 141: Be careful here, this will apply to every top-level slotted element.

brian-patrick-3 (changes requested):

src/components/reusable/modal/modal.scss
tulasigudibanda: Updated

src/components/reusable/modal/modal.stories.js
tulasigudibanda: Done

brian-patrick-3 (approved):

kumarivijeta (approved):

srpriyankashetty:
Hi @tulasigudibanda, cancel button boarder color - is this expected
image

srpriyankashetty:
Modal form text input box boarder when Ai enabled
image

tulasigudibanda:
> Hi @tulasigudibanda, cancel button boarder color - is this expected image

Ghost button doesn't have an AI variant.  @brian-patrick-3 , please confirm

tulasigudibanda:
> Modal form text input box boarder when Ai enabled image

There is no AI variant for text input in Figma. https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=4045-141810&p=f&m=dev

brian-patrick-3:
> > Hi @tulasigudibanda, cancel button boarder color - is this expected image
> 
> Ghost button doesn't have an AI variant.  @brian-patrick-3 , please confirm

That's right. And the form input example in AI mode can be ignored.

Issue #386, opened 2025-02-21
Card: Highlight State


brian-patrick-3:
### Describe the bug

The Card component is missing the Highlight/Selected state as depicted in the [Vital Card pattern in Figma](https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=9944-1883&m=dev).

[Original report in Teams](https://teams.microsoft.com/l/message/19:m15yXA3kC7fl9TvnIML-6DxhfDtwyoLJMu0nVAeddXQ1@thread.tacv2/1740039532682?tenantId=f260df36-bc43-424c-8f44-c85226657b01&groupId=6097bcb7-eb77-4588-9226-1430b9c803ab&parentMessageId=1740039532682&teamName=Shidoka%20Design%20System&channelName=General&createdTime=1740039532682)

[AB#2124752](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/2124752)

### To do

- [ ] Add highlight prop to Card component
- [ ] Update Notification component "unread" state to use the new Card prop

brian-patrick-3:
Will address after #363 is merged

PR #396, opened 2025-02-26
fix(card): card highlight fix

kumarivijeta:
## Summary

Added a highlight prop for Card . Additionally, Notification component (type-clickable)  read/unread control using card highlight prop now.

## GitHub Issue Link

fixes https://github.com/kyndryl-design-system/shidoka-applications/issues/386

## Checklist

### DIFF
--- a/src/components/reusable/notification/notification.ts
+++ b/src/components/reusable/notification/notification.ts
@@ -136,6 +136,7 @@ export class Notification extends LitElement {
             class="${this.unRead
               ? 'notification-mark-unread'
               : 'notification-mark-read'}"
+            ?highlight=${this.unRead}
             type=${this.type}
             href=${this.href}
             target="_blank"
--- a/src/components/reusable/notification/notification.ts
+++ b/src/components/reusable/notification/notification.ts
@@ -133,7 +133,7 @@ export class Notification extends LitElement {
     return html`
       ${this.type === 'clickable'
         ? html`<kyn-card
-            class="notification-mark-unread"
+            class="notification-clickable"
             ?highlight=${this.unRead}
             type=${this.type}
             href=${this.href}
--- a/src/components/reusable/notification/notification.ts
+++ b/src/components/reusable/notification/notification.ts
@@ -133,9 +133,7 @@ export class Notification extends LitElement {
     return html`
       ${this.type === 'clickable'
         ? html`<kyn-card
-            class="${this.unRead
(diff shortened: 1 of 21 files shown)


src/components/reusable/notification/notification.ts
```diff
@@ -136,6 +136,7 @@ export class Notification extends LitElement {
             class="${this.unRead
               ? 'notification-mark-unread'
               : 'notification-mark-read'}"
```
brian-patrick-3, line 138: Are these classes still needed?

kumarivijeta: Yes, either one is required to get the text visibility in dark mode when type 'clickable' & unread is 'true'.

srpriyankashetty (approved):
Verified the Card highlight control for clickable and normal type.
Verified the Notification highlight works fine for clickable.

src/components/reusable/notification/notification.ts
brian-patrick-3: Shouldn't the card highlight styles drive the text color, not the notification?

kumarivijeta: You mean when card is highlighted , the text color should be in green color ?  

The Fix for text color style is adjusted due to readbility issue in dark mode specially on notification.

image

brian-patrick-3: Designers should provide a text color to go with the card background color.  I think they skipped that here because its a slot

kumarivijeta: Robert confirmed that, The card background color shouldn't control the text color. The content within the card is placeholder or slotted, and the application should handle the styling as needed.

brian-patrick-3: I have to disagree. Text color is inherited, and needs to contrast the background. When we're changing the background color in a container like this, we also need to set a usable text color. The content in the slot will inherit that color.

kumarivijeta: I have applied a color to the slot, allowing the content inside the slot to inherit that color. This works perfectly for Cards. However, in the Notification component, the text color is being overridden for Toast/Inline Messages because the background color remains consistent across both light and dark modes. 
Since, we are using the same HTML template and styles for both  notifications type 'normal' and 'clickable', the overridden color style causes a contrast problem. To resolve this, I added additional styles.

brian-patrick-3 (approved):

PR #402, opened 2025-02-28
feat(inline-confirm): add reusable component inline-confirm

brian-patrick-3:
## Summary

Add a reusable inline confirmation component in support of the AI Info & Chat History patterns.

fixes AB#2136681

## Figma Link

https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=29700-3508&m=dev

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -1691,69 +1691,6 @@
     },
     {
       "kind": "javascript-module",
-      "path": "src/components/global/uiShell/index.ts",
-      "declarations": [],
-      "exports": [
-        {
-          "kind": "js",
-          "name": "UiShell",
-          "declaration": {
-            "name": "UiShell",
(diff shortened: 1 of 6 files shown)


tulasigudibanda (approved):

tulasigudibanda (approved):

wbarbee (approved):
i assume that this is intended, but just confirming that this should always be offset a bit to the right (not centered)? looks like it's leaving space for the animation and icon position shift.

brian-patrick-3:
> i assume that this is intended, but just confirming that this should always be offset a bit to the right (not centered)? looks like it's leaving space for the animation and icon position shift.

Exactly yes, the latest revision reserves space for the hidden buttons to prevent overlap

PR #410, opened 2025-03-04
fix(pickers): _handleClear logic issue, date, time, date-range picker component refactor

### LINKED ISSUE #409: handleClear not working as expected in date-range (date, time) pickers
**Before submitting a bug:**

- [x] Is this reproducible in the latest version of Shidoka?
- [x] Is this reproducible in the [Shidoka Storybook](https://shidoka-applications.netlify.app)?
- [ ] Is this reproducible in a [Sandbox](https://codesandbox.io)?
- If it's not reproducible in any of the above scenarios, please do not open a bug report as it is an application-level issue.

**Describe the bug**

Dev reports:
"I've tried multiple possible solutions, but I cannot figure why the date range picker values do not clear when the values attribute is set to [null, null] 
 
Also, how can I capture the clear event, when I click the 'x' in the input?"

**Expected behavior**

Clicking the 'X' button after values are selected in the date-range, date, time pickers should emit an event, it should emit an action in Storybook and it should clear the values attribute.

**Environment**

- OS: MacOS 15.3.1 (24D70)
- Browser: Chrome Version 133.0.6943.127 (Official Build) (arm64)
- Version: 2.8.2

[AB#2143307](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/2143307)

wbarbee:
## Summary

- fix defect that was preventing clear event from firing in certain scenarios -- improve _handleClear logic for all three pickers
- enhance date, time, date-range picker Storybook files with on-change action firing on X-icon clear button click for sanity check
- refactored all three components (see Notes section below for details)

fixes #409 
fixes [AB#2143307](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/2143307)

## Checklist

#### Flatpickr helper file
- consistent date format handling
- simplified input element handling
- removed redundant code

#### Timepicker
- better use of lifecycle methods, improved event handling and timing 

#### DatePicker
- Ensured that the placeholder value is generated using the existing getPlaceholder function.
- Standardized the handling of the disable prop by mapping any string dates to Date objects when possible.
- Made sure that updates to properties (like defaultDate, dateFormat, and others) trigger the appropriate reinitialization or update of the flatpickr instance.

#### Date-Range Picker
- added additional `_defaultTextStrings`
- ensure recent reported issues are addressed:
1. Dynamic positioning is achieved by detecting if the component is within a modal (using `this.closest('kyn-modal')`) and setting the appendTo container accordingly. The setCalendarAttributes calls pass a flag to adjust positioning.

2. The validation logic in `_validate` ensures that if the component is required but no dates are selected (i.e. [null, null]), a non‑empty validation message is always provided. This prevents the setValidity error when required fields are empty.

3. Adjusted `_handleClear` clear functionality so that when the clear button is clicked or the value is set to [null, null], the flatpickr instance is cleared and the input’s value is reset. Additionally, an on-change event is emitted with a source: 'clear' payload, allowing consumers to capture the clear event.

## Testing Instructions

In the context of Storybook on-clear action:
1. Go to Date Range Picker > Date Range Default
2. Select a date range to populate the input field
3. Click the 'X' to clear the input and note the updated on-clear action that is available in the Actions tab


brian-patrick-3:
I would suggest emitting the on-change event on clear since it does change the value, on-clear may not be necessary.

brian-patrick-3 (approved):

brian-patrick-3:
I actually just had one thought. That logic we have to check for the kyn-modal, we should probably also check for kyn-side-drawer.

wbarbee:
> I actually just had one thought. That logic we have to check for the kyn-modal, we should probably also check for kyn-side-drawer.

a good thought. we can have an array of components to check against as we continue to build more. i'll get logic in there. thanks

srpriyankashetty (approved):
Verified the Date range picker, Date picker and Time picker looks fine

@wbarbee, Please consider there for enhancement
1. Mobile view the date picker must be adjusted
2. Calendar icon clicking 3rd time does not open the date picker
3. Multi select scroll option to see the dates or tooltip to show the selected dates
4. Either remove the twentyfourhour format props or remove the H and h format - keeping both is confusing.
5. Self-adjusting (automatic positioning) date picker when Modal pop-up is open, or side drawer is open

PR #421, opened 2025-03-11
fix(pickers): inline-block preventing parent width + date/timepicker value

wbarbee:
## Summary

1. Dev reported that pickers were not taking up full width of parent container as expected/intended.

2. improve value vs. defaultValue logic


src/common/scss/mixins.scss
```diff
@@ -21,6 +21,7 @@
 @mixin picker-base {
   :host {
     display: inline-block;
+    width: 100%;
```
brian-patrick-3, line 24: All of our inputs are inline-block/auto width by default. The application can make them full width using CSS on the host tag.

brian-patrick-3 (approved):

tulasigudibanda (approved):

PR #422, opened 2025-03-11
feat(ai-prompts, ai-answers): update prompts and answers

wbarbee:
## Summary

After a talk with Matthew, there was a feature request to see how the AI Prompts and Answers might be animated on exit when selected.

~This PR adds a smooth, animated effect that briefly accentuates the selected item and then fades all items out of view.~ **Archive animation-related code -- to potentially be integrated and componentized at a later date**

## Additional
Screenshot 2025-03-11 at 4 45 58 PM

<br/>
*** the customization to remove the sample card logo and subtitle (neither of which are utilized in this AI implementation of the clickable card) was unreliable --> added some card props to the `card.sample.ts` to make that sample component a bit more customizable.

### DIFF
--- a/src/stories/ai/aiPrompts.stories.js
+++ b/src/stories/ai/aiPrompts.stories.js
@@ -1,5 +1,6 @@
 import { html } from 'lit';
 import { action } from '@storybook/addon-actions';
+
 import '../../components/reusable/card';
 
 export default {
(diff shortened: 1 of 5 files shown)


brian-patrick-3:
I do like this animation, but I'm not sure about having it in the format of a pattern, because we can't really serve it up to applications this way, and if we make changes they will not cascade. It is also left open to interpretation/modification. If we want to truly provide this animation for application consumption, I think we would have to compenentize the Prompts. This is how we ended up with the Inline Confirm component for example, componentized it so we could serve the animations.

wbarbee:
> I do like this animation, but I'm not sure about having it in the format of a pattern, because we can't really serve it up to applications this way, and if we make changes they will not cascade. It is also left open to interpretation/modification. If we want to truly provide this animation for application consumption, I think we would have to compenentize the Prompts. This is how we ended up with the Inline Confirm component for example, componentized it so we could serve the animations.


That all makes sense. This was a request late in the afternoon, but a good one. So I guess this can serve as sort of proposed animation we can show to Matthew and discuss if it's worth converting it to a component now, later, or leaving it as it currently is on main. Thanks.

brian-patrick-3:
Yeah exactly. This is good stuff and I want more, it just has to be thoughtfully designed and included as part of our components vs tacked on.

wbarbee:
@brian-patrick-3 i've archived the code related to the animation effects, so this PR just fixes a bug where the card-logo and card subtitle were inconsistently removed from the simplified clickable card.

i'd still like to merge this one in

src/stories/ai/aiPrompts.stories.js
```diff
@@ -12,177 +13,284 @@ export default {
...
+      </style>
+
+      <div class="ai-prompts-wrapper">
+        <kyn-card
+          type="clickable"
+          aiConnected
+          href="javascript:void(0)"
+          @on-card-click=${handleCardClick}
+        >
+          <sample-card-component .showLogo=${false} .subtitle=${null}>
```
brian-patrick-3, line 106: We probably shouldn't be using the sample card component here. I see some CSS styles that are tied to using that, but applications will not use that so it's not going to translate over.

src/stories/ai/aiPrompts.stories.js
```diff
@@ -12,177 +13,284 @@ export default {
...
+        .ai-prompts-wrapper {
+          display: flex;
+          gap: 20px;
+          flex-wrap: wrap;
+        }
+
+        .ai-prompts-wrapper kyn-card {
+          width: 250px;
+          font-size: 14px;
+          transition: transform 0.4s ease, opacity 0.5s ease;
```
brian-patrick-3, line 55: Should also archive the animation related CSS.

src/stories/ai/aiPrompts.stories.js
```diff
@@ -12,177 +13,284 @@ export default {
...
+      //   allCards.forEach((c) => {
+      //     c.classList.add('fade-out');
+      //   });
+      // }, 300);
+
+      action(e.type)(e);
+    };
+
+    return html`
+      <style>
```
brian-patrick-3, line 184: Can any of this be re-used for both stories to be less redundant?

brian-patrick-3 (approved):

jaideep08 (approved):

srpriyankashetty:
@wbarbee, where can I test these changes and can you please provide the test instructions.!
I don't see the screenshot mentioned above in the AI prompt

wbarbee:
> @wbarbee, where can I test these changes and can you please provide the test instructions.! I don't see the screenshot mentioned above in the AI prompt

@srpriyankashetty this was really just a code enhancement to ensure the card-logo (circular image) and subtitle didn't show on load sometimes. a light refactor, which is why i went ahead and merged it.

i've included screenshots above and these patterns are observable here:
- [AI Answers](https://shidoka-applications.netlify.app/?path=/docs/ai-patterns-ai-answers--docs&args=darkTheme:light)
- [AI Prompts](https://shidoka-applications.netlify.app/?path=/story/ai-patterns-ai-prompts--default)

PR #424, opened 2025-03-12
fix: gen ai styles audit

brian-patrick-3:
Focus on fixing any colors that look broken and getting the tests to pass in this PR. More complicated changes like new Tabs styles can be part of next release.

### List of manual changes from the design team:

- [x] Avatar : Background color 
- [x] Card container: border, background
- [x] Data Table: background colors, header rounded corners added (4px), alternate color background update
- [x] Global Navigation: lots of changes happened here:
  - Background color updates to all variants
  - Added a new SM size, and moved the other ones up one size, we now have LG, MD, SM. SM is for cases when they have a lot of items in a menu and they’re categorized, need less padding
  - Category variant has additional padding above it for when multiple categories are show there is more of a dividing factor to it
- [x] Link: changed color back to spruce greens
- [x] Modal and Side drawer, I think these continue to use the Background Container Default, but now show lighter color
- [x] Tabs: lots of updates happened to this component as well
  - Consolidated all the colors same as Buttons, and brought back the spruce for the tab component non-ai
  - All states got updated, notice the details as well like the padding was removed on the sides so there is no gap on the line that connects all the tabs
- [x] Accordion: some color changes happened to this as well, I think it was just a background color change so far

### Additional items found by dev team:

- [x] Button
- [x] Link - Secondary AI variant appears to be new

### Token issues that need to be resolved in Figma:
- [x] Link - Primary color fails contrast


src/components/reusable/table/table-header.scss
```diff
@@ -9,6 +9,7 @@
   color: var(--kd-color-text-level-primary);
   font-weight: var(--kd-font-weight-medium);
   background: var(--kd-color-background-container-soft);
+  border-radius: 4px;
   border-right: 1px solid var(--kd-color-border-variants-inverse);
   border-left: 1px solid var(--kd-color-border-variants-inverse);
```
brian-patrick-3, line 14: @tulasigudibanda These borders appears to be poking out of the top left and right after adding the radius.

tulasigudibanda: Are you pointing to these @brian-patrick-3 :?
image

brian-patrick-3: > Are you pointing to these @brian-patrick-3 :?

Screenshot 2025-03-13 100647

src/components/reusable/accordion/accordion.scss
```diff
@@ -10,19 +10,19 @@
...
+    border-top: 1px solid var(--kd-color-border-level-secondary);
+    border-bottom: 1px solid var(--kd-color-border-level-secondary);
 
     ::slotted(kyn-accordion-item) {
-      border-top: 1px solid var(--kd-color-border-ui-default);
+      border-top: 1px solid var(--kd-color-border-level-secondary);
     }
 
     ::slotted(kyn-accordion-item:first-child) {
       border-top: none;
```
wbarbee, line 21: this is unrelated to these recent changes, but should the chevron be more vertically centered with the text here?
Screenshot 2025-03-13 at 10 09 31 AM

src/components/reusable/avatar/avatar.scss
```diff
@@ -6,11 +6,11 @@
   display: flex;
   align-items: center;
   justify-content: center;
-  width: 20px;
-  height: 20px;
+  width: 24px;
```
wbarbee, line 9: Screenshot 2025-03-13 at 10 10 09 AM

the size was increased, but the padding here is still a bit uneven between the text and container

wbarbee (commented):
two small things

src/components/reusable/accordion/accordion.scss
brian-patrick-3: SVG's are inline, which often adds some space below them by default. I usually use CSS `display: block;` to fix that.

src/components/reusable/avatar/avatar.scss
brian-patrick-3: That might just be the lettering giving the illusion of being uneven, hard to tell.

wbarbee (approved):

srpriyankashetty:
@brian-patrick-3, click or pressed action is not showing for this chat send button - is it fine?

image

srpriyankashetty:
@brian-patrick-3, Global navigation - second point related to size, is it in the backend?

srpriyankashetty:
Link in dark background shows accessibility error
Ensure the contrast between foreground and background colors meets WCAG 2 AA minimum contrast ratio thresholds

srpriyankashetty (commented):
Verified and created bugs for the issues observed

PR #428, opened 2025-03-21
feat: add readonly prop to all textual form controls

tulasigudibanda:
## Summary

Added readonly prop to the following form controls / web components :
Text Input
Text Area
Date Picker
Time Picker
Date Range Picker
Number Input

Difference between enabled and readonly states :  border and background colors.

[AB#2097596](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2097596)

## Figma Link

[Figma] (https://www.figma.com/design/rC5XdRnXVbDmu3vPN8tJ4q/2.1-Edinburgh?node-id=4099-10868&m=dev)

image

## To Do

Date Range Picker :
 1. [x] Date/Value getting reset when dateRangePickerDisabled is changed to true in storybook UI  - fixed in [PR/429](https://github.com/kyndryl-design-system/shidoka-applications/pull/429) 
Apply the same fix to readonly prop and inspect any other relevant inputs in Storybook for necessary updates.

2. [ ] Form validation incorrectly throws errors on submit for pickers with preselected selections.
https://github.com/user-attachments/assets/a936e50e-675f-4958-ab35-e74a7db91d69

3. Temporarily disabling or commenting out the isEmpty validation does not resolve the issue—formData is generated incorrectly since preselected values are not factored.
https://github.com/user-attachments/assets/8cceb48c-123b-4271-ba46-27053f1ebaa0

4. [ ] Clicking the picker input triggers an error message prompting the user to select options, even before a selection is attempted - Is it intended @brian-patrick-3 , @wbarbee ?
https://github.com/user-attachments/assets/1a8c6671-fc83-418e-87ae-02b1c656135d

5. [ ] In daterangepicker.stories.js, the below preselected values for WithPreselectedDateTime are defined in a non-ISO format, which the current implementation in daterangepicker.ts does not support.
   defaultDate: ['2024-01-01 09:00:00', '2024-01-02 17:00:00']
   
   wb: for this one, @tulasigudibanda, if you're setting these in the Storybook controls, ensure that the `dateFormat` is correct (ex: `Y-m-d h:i K`) to accept pre-selected date + time and if you change the single quotes to double quotes, it should work. we're handling this formatting when set to `['2024-01-01 09:00:00', '2024-01-02 17:00:00']` in the code itself.

6. [ ]  @brian-patrick-3 , @wbarbee , what should the error message be if the DateRangePicker has an incomplete preselected value (array length less than 2) and the required property is false?

Bug  for the above issues :  [Issues with daterangepicker](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2198863) 
Fix : https://github.com/kyndryl-design-system/shidoka-applications/pull/432

-----------------------------------------------------------------------------

Separate feature will be created for disabled state changes across all the applicable components


wbarbee:
1. i'm able to see the focus outline on the readonly date picker, timepicker, daterangepicker, text input, textarea -- is this desired?
2. Cursor (on hover) on the incrementing `<` and `>` of the numberInput and the readonly state in buttons should be default?
3. Calendar icons and Clock icons — I know the figma file shows these in read-only state, but should we? this communicates to me that the inputs are interactive
4. Not seeing any style differences for read-only vs. default state in buttons

tulasigudibanda:
> 1. i'm able to see the focus outline on the readonly date picker, timepicker, daterangepicker, text input, textarea -- is this desired?
> 2. Cursor (on hover) on the incrementing `<` and `>` of the numberInput and the readonly state in buttons should be default?
> 3. Calendar icons and Clock icons — I know the figma file shows these in read-only state, but should we? this communicates to me that the inputs are interactive
> 4. Not seeing any style differences for read-only vs. default state in buttons

@wbarbee 
1.  As per figma, readonly fields are focusible.
Screenshot 2025-03-28 at 2 27 08 PM
2. Changed cursor: not-allowed
3. Yes. it seems interactive. I checked this with Chandan earlier but she told me to keep it . 
4. readonly attribute is not relevant for button element . I had to add a prop on the button component as I was not able to pierce the kyn-button shadow DOM to add readonly styles.

src/components/reusable/button/button.ts
```diff
@@ -90,6 +90,10 @@ export class Button extends LitElement {
   @property({ type: Boolean, reflect: true })
   disabled = false;
 
+  /** Determines if the button is readonly. */
+  @property({ type: Boolean, reflect: true })
+  readonly = false;
```
brian-patrick-3, line 95: I don't think this is necessary on button, we can just use `disabled` for the buttons in a readonly number input.

brian-patrick-3 (changes requested):
Looks pretty good except for a couple accessibility errors and my comment about button.

brian-patrick-3 (approved):

srpriyankashetty:
@tulasigudibanda, is this case implemented?
image

srpriyankashetty (approved):
Verified the following components read only mode and looks fine
Verified the background color when field is read only in both dark and light mode - matches with Figma
Verified the Focus and select using keyboard action when read only is selected
Verified that text selection is allowed when made read only

Text Input
Text Area
Date Picker
Time Picker
Date Range Picker
Number Input

wbarbee (approved):

PR #432, opened 2025-03-25
fix(datepicker, timepicker, daterangepicker): pre-selected, defaultDate handling

wbarbee:
## Summary

When utilizing Flatpickr's options `defaultDate` config (`defaultHour` and `defaultMinute` for timepicker) to set a pre-selected input date/time/date-time, these values were not being passed along with the rest of the final, submitted form data.

This PR refactors specifically how `defaultDate`, `defaultHour`, `defaultMinute` is being handled in the absence of user-selected `this.value`.

Screenshot 2025-03-25 at 11 10 55 AM

AB#2198863

## Notes

- **BREAKING CHANGE**: removes the `defaultDate` prop from timepicker.ts to avoid confusion -- only `defaultHour` and `defaultMinute` will be used going forward.

- [x]  Date/Value getting reset when dateRangePickerDisabled is changed to true in storybook UI  - fixed in [PR/429](https://github.com/kyndryl-design-system/shidoka-applications/pull/429)

- [x] Form validation incorrectly throws errors on submit for pickers with preselected selections.
Attached the recording 

- [x] Temporarily disabling or commenting out the isEmpty validation does not resolve the issue—formData is generated incorrectly since preselected values are not factored.
Attached the recording 

- [ ] Clicking the picker input triggers an error message prompting the user to select options, even before a selection is attempted - Is it intended ?
Attached the recording 

- [x] In daterangepicker.stories.js, the below preselected values for WithPreselectedDateTime are defined in a non-ISO format, which the current implementation in daterangepicker.ts does not support.
   defaultDate: ['2024-01-01 09:00:00', '2024-01-02 17:00:00']

## Checklist

## Testing Instructions

#### Illustrates that pre-selected time, date, multi-date, and date-time ranges are all being successfully submitted now:
Screenshot 2025-03-25 at 11 24 33 AM

Note: `(2) ['date-time-range-1', 'Wed Jan 01 2025 18:00:00 GMT-0600 (Central Standar…12 2025 18:00:00 GMT-0600 (Central Standard Time)']`

### DIFF
--- a/src/components/reusable/datePicker/datepicker.ts
+++ b/src/components/reusable/datePicker/datepicker.ts
@@ -192,6 +192,10 @@ export class DatePicker extends FormMixin(LitElement) {
   @state()
   private _shouldFlatpickrOpen = false;
 
+  /** Track if we initially had a defaultDate when the component was first connected */
+  @state()
+  private _hasInitialDefaultDate = false;
+
   /** Track initialization state
    * @internal
    */
@@ -238,9 +242,6 @@ export class DatePicker extends FormMixin(LitElement) {
       this._hasInitialDefaultDate = true;
     }
   }
-
-  @state()
-  private _hasInitialDefaultDate = false;
 
   private hasValue(): boolean {
     if (this._inputEl?.value) return true;
--- a/src/components/reusable/datePicker/datepicker.ts
+++ b/src/components/reusable/datePicker/datepicker.ts
@@ -233,7 +233,14 @@ export class DatePicker extends FormMixin(LitElement) {
     super.connectedCallback();
     this.addEventListener('change', this._onChange);
     this.addEventListener('reset', this._handleFormReset);
+
+    if (!this.value && this.defaultDate) {
+      this._hasInitialDefaultDate = true;
+    }
   }
+
+  @state()
+  private _hasInitialDefaultDate = false;
 
   private hasValue(): boolean {
     if (this._inputEl?.value) return true;
@@ -394,7 +401,43 @@ export class DatePicker extends FormMixin(LitElement) {
       this._initialized = true;
       await this.updateComplete;
       this.setupAnchor();
+
+      if (this._hasInitialDefaultDate && this.defaultDate) {
+        const processedDates = this.processDefaultDates(this.defaultDate);
+        if (processedDates && processedDates.length > 0) {
+          if (this.mode === 'multiple') {
+            this.value = [...processedDates];
+          } else {
+            this.value = processedDates[0];
+          }
+
+          this._internals.setFormValue(this.value);
+        }
+      }
     }
+  }
+
+  private processDefaultDates(
+    defaultDate: string | string[] | Date | Date[] | null
+  ): Date[] {
+    if (!defaultDate) return [];
+
+    if (Array.isArray(defaultDate)) {
+      return defaultDate
+        .map((date) => {
+          if (date instanceof Date) return date;
+          if (typeof date === 'string') return this.parseDateString(date);
+          return null;
+        })
+        .filter((date): date is Date => date !== null);
+    } else if (typeof defaultDate === 'string') {
+      const parsed = this.parseDateString(defaultDate);
+      return parsed ? [parsed] : [];
+    } else if (defaultDate instanceof Date) {
+      return [defaultDate];
+    }
+
+    return [];
   }
 
   override updated(changedProperties: PropertyValues) {
@@ -602,6 +645,31 @@ export class DatePicker extends FormMixin(LitElement) {
     await this.debouncedUpdate();
   }
 
+  private parseDateString(dateStr: string): Date | null {
+    if (!dateStr || !dateStr.trim()) return null;
+
+    if (dateStr.includes('T')) {
+      const date = new Date(dateStr);
+      return isNaN(date.getTime()) ? null : date;
+    }
+
+    const formats: { [key: string]: RegExp } = {
+      'Y-m-d': /^\d{4}-\d{2}-\d{2}$/,
+      'Y-m-d h:i K': /^\d{4}-\d{2}-\d{2}( \d{1,2}:\d{2} [AP]M)?$/,
+    };
+
+    const pattern = formats[this.dateFormat];
+    if (!pattern || !pattern.test(dateStr)) return null;
+
+    const [datePart] = dateStr.split(' ');
+    const [year, month, day] = datePart.split('-').map(Number);
+
+    if (!year || !month || !day) return null;
+
+    const date = new Date(year, month - 1, day);
+    return isNaN(date.getTime()) ? null : date;
+  }
+
   setInitialDates(): void {
     if (!this.flatpickrInstance) {
       console.warn(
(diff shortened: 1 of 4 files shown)


tulasigudibanda:
@wbarbee , when required=true and switch dateRangePickerDisabled=true in storybook , error-text is rendered . 
https://github.com/user-attachments/assets/ff520550-5f5b-46d7-8592-20efadbdb811

tulasigudibanda:
preselected defaultDate in WithPreselectedDateTime story needs to updated to ISO format if non ISO values as below are not going to be supported : (formData is generated as empty for non ISO values)
[
0 : "2024-01-01 09:00:00"
1 : "2024-01-02 17:00:00"
]
image

wbarbee:
> preselected defaultDate in WithPreselectedDateTime story needs to updated to ISO format if non ISO values as below are not going to be supported

i think the format would be: 
`[
"2024-01-01 09:00:00",
"2024-01-02 17:00:00"
]`

and it should work as expected.

the error showing when `required` and `dateRangePickerDisabled` are both true should be fixed in this latest commit

wbarbee:
> > > preselected defaultDate in WithPreselectedDateTime story needs to updated to ISO format if non ISO values as below are not going to be supported
> > 
> > 
> > i think the format would be: `[ "2024-01-01 09:00:00", "2024-01-02 17:00:00" ]`
> > and it should work as expected.
> 
> @wbarbee , although the date-time range appears to be displayed correctly , it does not actually generate formData upon submission. https://github.com/user-attachments/assets/88936b13-bad4-405a-97d8-7d32132e7c6c
> 
> > the error showing when `required` and `dateRangePickerDisabled` are both true should be fixed in this latest commit
> > Fixed

ok, this latest commit should address this. i have the dates array `[ "2024-01-01 09:00:00", "2024-01-02 17:00:00" ]` loaded into the last pre-selected form input for easy confirmation:

i was able to reproduce the issue. this is the new output:
Screenshot 2025-03-26 at 8 30 03 AM

thanks @tulasigudibanda

tulasigudibanda:
@wbarbee , there is an inconsistency in behavior between the date range picker and the date + time range picker when only one value is selected. The date range picker resets to an empty value, whereas the date + time range picker defaults to the current date. This could lead to unintended time selections if the user doesn’t notice the change. Ideally, the behavior should be consistent—either resetting to an empty value or retaining the selected value and displaying an error.
https://github.com/user-attachments/assets/c1b13b61-abf5-4483-bbed-41b7af657eb1

wbarbee:
> @wbarbee , there is an inconsistency in behavior between the date range picker and the date + time range picker when only one value is selected. The date range picker resets to an empty value, whereas the date + time range picker defaults to the current date. This could lead to unintended time selections if the user doesn’t notice the change. Ideally, the behavior should be consistent—either resetting to an empty value or retaining the selected value and displaying an error. https://github.com/user-attachments/assets/c1b13b61-abf5-4483-bbed-41b7af657eb1

yes, this occurs because we're defaulting the Flatpickr config option `closeOnSelect` to `true` **unless** a time component is present in the date/date-range picker. i don't know of a clear solution to this one. the calendar UI needs to remain open to allow for date and time selection, otherwise, it will just close after two dates are selected without giving the user the chance to select time(s) without re-opening the calendar UI.

tulasigudibanda:
> closeOnSelect



> > @wbarbee , there is an inconsistency in behavior between the date range picker and the date + time range picker when only one value is selected. The date range picker resets to an empty value, whereas the date + time range picker defaults to the current date. This could lead to unintended time selections if the user doesn’t notice the change. Ideally, the behavior should be consistent—either resetting to an empty value or retaining the selected value and displaying an error. https://github.com/user-attachments/assets/c1b13b61-abf5-4483-bbed-41b7af657eb1
> 
> yes, this occurs because we're defaulting the Flatpickr config option `closeOnSelect` to `true` **unless** a time component is present in the date/date-range picker. i don't know of a clear solution to this one. the calendar UI needs to remain open to allow for date and time selection, otherwise, it will just close after two dates are selected without giving the user the chance to select time(s) without re-opening the calendar UI.

Can we at least reset to an empty value when only one value is selected in the date range picker?

tulasigudibanda:
> > closeOnSelect
> 
> > > @wbarbee , there is an inconsistency in behavior between the date range picker and the date + time range picker when only one value is selected. The date range picker resets to an empty value, whereas the date + time range picker defaults to the current date. This could lead to unintended time selections if the user doesn’t notice the change. Ideally, the behavior should be consistent—either resetting to an empty value or retaining the selected value and displaying an error. https://github.com/user-attachments/assets/c1b13b61-abf5-4483-bbed-41b7af657eb1
> > 
> > 
> > yes, this occurs because we're defaulting the Flatpickr config option `closeOnSelect` to `true` **unless** a time component is present in the date/date-range picker. i don't know of a clear solution to this one. the calendar UI needs to remain open to allow for date and time selection, otherwise, it will just close after two dates are selected without giving the user the chance to select time(s) without re-opening the calendar UI.
> 
> Can we at least reset to an empty value when only one value is selected in the date+time range picker , if retaining the selection is not possible ? That way user can select both values again.

Tested the fix and is getting reset to empty value consistently

src/components/reusable/datePicker/datepicker.ts
```diff
@@ -615,19 +699,47 @@ export class DatePicker extends FormMixin(LitElement) {
...
+          }
         }
-      } else if (typeof dateToSet === 'string' && dateToSet.trim()) {
-        this.flatpickrInstance.setDate(dateToSet, true);
       } else if (dateToSet instanceof Date) {
         this.flatpickrInstance.setDate(dateToSet, true);
+
+        if (this.value === null) {
+          this.value = dateToSet;
+          this._validate(false, false);
```
brian-patrick-3, line 741: I think these `_validate` calls are also redundant, as validation also triggers when `value` updates.

brian-patrick-3 (approved):

tulasigudibanda (approved):

tulasigudibanda:
@wbarbee  , would it be possible to remove the
 onload warnings(Cannot set initial dates: Flatpickr instance not available) 
and 
intermediate console log statements to reduce confusion, keeping only the necessary logs when the picker input loses focus? 

https://github.com/user-attachments/assets/5a0ef9c3-bc1d-437f-b2db-13d2db72be0c

wbarbee:
> @wbarbee, In the following image,
> 
> 1. the warning message is not showing when there is required error message displayed.
> 2. What is this value field, as we already have default date, is this required and does not show the value on passing the date.
>    Please confirm on these.
> 
> image

1. for the warning, that is behaving as expected. the error message should take precedence over the warning.

2. I agree this is confusing. ideally, we would convert `this.value` to a state variable `this._value` and make it internal to each of the components, however that would introduce a more impactful breaking change than the one that we are currently introducing.

I've improved the JSDOC documentation for clarity and removed the `value` control from the storybook files to reduce confusion.

wbarbee:
@brian-patrick-3 hopefully last thing on this. 

we have this `value` property that is really only used internal to the component when a user selects a date/time value. ideally, we would convert this to a state property to reduce confusion and introduce another breaking change. for now, i've removed the value control from the Storybook file to reduce confusion, but if the consensus is that it would be preferable to just convert it to a state variable while we're at it, i can do that in this PR.

brian-patrick-3:
> @brian-patrick-3 hopefully last thing on this.
> 
> we have this `value` property that is really only used internal to the component when a user selects a date/time value. ideally, we would convert this to a state property to reduce confusion and introduce another breaking change. for now, i've removed the value control from the Storybook file to reduce confusion, but if the consensus is that it would be preferable to just convert it to a state variable while we're at it, i can do that in this PR.

Hmm for every other form control using the form mixin, `value` is the prop that the app uses to control the value, and is what gets submitted with the form. The datepickers seem to be the oddball here with that pattern. If they're overriding that behavior and we can just use `@state` and `@internal` without it breaking, that's fine I suppose.

wbarbee:
> > @brian-patrick-3 hopefully last thing on this.
> > we have this `value` property that is really only used internal to the component when a user selects a date/time value. ideally, we would convert this to a state property to reduce confusion and introduce another breaking change. for now, i've removed the value control from the Storybook file to reduce confusion, but if the consensus is that it would be preferable to just convert it to a state variable while we're at it, i can do that in this PR.
> 
> Hmm for every other form control using the form mixin, `value` is the prop that the app uses to control the value, and is what gets submitted with the form. The datepickers seem to be the oddball here with that pattern. If they're overriding that behavior and we can just use `@state` and `@internal` without it breaking, that's fine I suppose.

Flatpickr has a native `defaultDate` value that allows the user to set the initial date. i think that in distinguishing defaultDate from the value property, some confusion was introduced. converting `value` to an internal `_value` shouldn't break anything, but i wanted to check. i assume i should still mention it in the breaking changes?

brian-patrick-3:
> Flatpickr has a native `defaultDate` value that allows the user to set the initial date. i think that in distinguishing defaultDate from the value property, some confusion was introduced. converting `value` to an internal `_value` shouldn't break anything, but i wanted to check. i assume i should still mention it in the breaking changes?

I think you'll still need it to be `value` for the FormMixin, but simply marking it as `@state` or `@internal` should be fine.

wbarbee:
@srpriyankashetty -- i made a simple conversion from @property to @state/@internal for `value`. i've tested and everything looks good. but, as discussed, if you could please give it one more run-through to make sure everything looks good on your end. thank you!

srpriyankashetty (approved):
Verified the all the issues mentioned in the Notes section
Verified all the scenarios mentioned in the Test instructions
Verified the Required and disable related issue and looks fine
Verified the **Value** field removed from the control
Verified the removal of **default date** from time picker
Verified that all the values are showing in console tab on submitting the Form 
Verified the min max dates, required toggle, caption, invalid, warning 

Observation - 
Date picker in Modal or side drawer, before opening the pop-up, set the date as required, add default date and then opened the modal pop-up. 
The required error message is showing even when the default date is present.
Rare scenario but mentioned to have it as a record.
CC: @wbarbee, @brian-patrick-3
image

wbarbee:
thank you @srpriyankashetty -- the observed edge case occurs in all scenarios, regular form view, modal, and side drawer.

it's unlikely that a user filling out a form would have the `required` and `defaultDate` inputs exposed to them. these properties would be set by the developer prior to the page load. 

thanks for documenting and we will watch and hope that this resolves the issues with these three components. thanks again for the diligence. merging in now.

PR #433, opened 2025-03-26
feat: update textarea in feedback form

tulasigudibanda:
Reduced height of feedback textarea to 3 rows
Allowed textarea resize.

[AB#2201491](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2201491)

## Figma Link

[Figma Link
](https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=26610-251300&p=f&m=dev)

### DIFF
--- a/src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
+++ b/src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
@@ -168,9 +168,8 @@ const feedbackFormContent = () => html`
     <kyn-text-area
       aiConnected
       class="input-text-area"
-      rows="8"
+      rows="3"
       placeholder="Provide additional feedback"
-      ?notResizeable=${true}
     ></kyn-text-area>
 
     <div class="footer">


kumarivijeta (approved):

jaideep08 (approved):

wbarbee (approved):

jaideep08 (approved):

srpriyankashetty:
@tulasigudibanda 
Show more/show less button text color in dark mode is not visible on hover.

image

tulasigudibanda:
> @tulasigudibanda
> 
> 1. Show more/show less button text color in dark mode is not visible on hover.
> 2. Observed accessibility violations in dark mode

'Show more' button text color is --kd-color-background-button-primary-ai-default  in figma which is not the default. Hence the issue.
Second issue is with the kyn-link inside the cards is throwing color contrast violation in dark mode
 I am waiting for Robert to check and get back on both of these.

src/components/ai/sourcesFeedback/aiSourcesFeedback.scss
```diff
@@ -16,12 +16,12 @@
 
 .card-container {
   display: grid;
-  gap: 16px;
+  gap: 0.5rem;
```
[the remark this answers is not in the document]
tulasigudibanda: @wbarbee , @brian-patrick-3 , isn't rem generally better for accessibility and scalability across different browsers?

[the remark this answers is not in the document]
brian-patrick-3: I actually usually use px for spacing, so spacing stays fixed when zooming. rem would scale with zoom though, which technically might be more accurate.

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
```diff
@@ -123,10 +117,18 @@ const SourcesContent = () => html`
         rel=${args.rel}
         ?hideBorder=${args.hideBorder}
       >
-        <h1 class="card-title">
-          <div>${card.title}</div>
-        </h1>
         <div class="card-description">
+          <kyn-button
```
brian-patrick-3, line 121: What is this button supposed to do? Should it just be an icon instead? The vertical line looks strange inside it too.

brian-patrick-3 (changes requested):
1. The padding on the cards doesn't match the designs. Need to reconcile that either by updating the designs, or reducing card padding in code here.
2. Should the cards and the links inside use the AI variants?
3. Cards are not stretching to fill empty space on small screens.
4. The "Show more/less" functionality/button, do you think we should keep that or let the application add/control it as needed? I ask because we really can't predict what they are injecting in those slotted cards.

tulasigudibanda:
> 1. The padding on the cards doesn't match the designs. Need to reconcile that either by updating the designs, or reducing card padding in code here.
> 2. Should the cards and the links inside use the AI variants?
> 3. Cards are not stretching to fill empty space on small screens.
> 4. The "Show more/less" functionality/button, do you think we should keep that or let the application add/control it as needed? I ask because we really can't predict what they are injecting in those slotted cards.

1,3 - done
2. Waiting for Natalia to confirm.
4. Waiting for Natalia to confirm on this: 
Are sources always  kyn-cards or can they even be deeply nested structures ? 
Because 'Show more'  logic assumes a uniform layout and countable items — which would break for deeply nested structures with how items are selected/hidden .

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
tulasigudibanda: Changed this to an icon. But waiting for Natalia to confirm the clickable items :
just the Source Link   
OR
Earth icon/btn +  Source Link
OR
Whole card ?

Added separator instead of border.

brian-patrick-3:
> 1,3 - done 
> 2. Waiting for Natalia to confirm.
> 4. Waiting for Natalia to confirm on this: Are sources always kyn-cards or can they even be deeply nested structures ? Because 'Show more' logic assumes a uniform layout and countable items — which would break for deeply nested structures with how items are selected/hidden .

Shouldn't it be our designers (Robert) confirming these items?

Slotted sources should be kyn-cards yes, but if the application does anything out of the ordinary like using extra divs, creating a wrapper component, or putting in really long links for example, it could mess with the structure if we're still controlling the number of visible cards. Just something to consider, it probably will be fine as-is.

tulasigudibanda:
> > 1,3 - done
> > 2. Waiting for Natalia to confirm.
> > 4. Waiting for Natalia to confirm on this: Are sources always kyn-cards or can they even be deeply nested structures ? Because 'Show more' logic assumes a uniform layout and countable items — which would break for deeply nested structures with how items are selected/hidden .
> 
> Shouldn't it be our designers (Robert) confirming these items?
> 
> Slotted sources should be kyn-cards yes, but if the application does anything out of the ordinary like using extra divs, creating a wrapper component, or putting in really long links for example, it could mess with the structure if we're still controlling the number of visible cards. Just something to consider, it probably will be fine as-is.

Previously Robert told me to confirm about earth icon with Natalia . So posted it in the group chat (with Robert and Chandan) . Sure. I will keep the existing logic in that case unless Natalia gets back with something different

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
```diff
@@ -135,11 +128,38 @@ const SourcesContent = () => html`
     `
   )}
   <style>
-    .card-title div {
-      @include typography.type-ui-01;
-      color: var(--kd-color-text-level-primary);
-      font-size: 16px;
-      font-weight: 500;
+    kyn-card {
+      width: 100%;
```
brian-patrick-3, line 132: This could also be an internal ::slotted style

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
```diff
@@ -135,11 +128,38 @@ const SourcesContent = () => html`
...
+    kyn-card::part(card-wrapper) {
+      padding: 8px 4px;
+    }
+    .card-description {
+      display: flex;
+      align-items: center;
+      .source-icon,
+      kyn-link {
+        padding: 0px 8px;
+      }
```
brian-patrick-3, line 143: Maybe use `gap` here instead so you don't have to add padding to individual flex items like this link and the globe icon.

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
```diff
@@ -135,11 +128,38 @@ const SourcesContent = () => html`
...
+    }
+
+    .source-icon {
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      opacity: 0.8;
+    }
+
+    .separator {
```
brian-patrick-3, line 153: You might need to add `flex-shrink: 0;` here. I notice sometimes the lines vanish when I adjust my window size.

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
```diff
@@ -135,11 +128,38 @@ const SourcesContent = () => html`
...
+      kyn-link {
+        padding: 0px 8px;
+      }
+    }
+
+    .source-icon {
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      opacity: 0.8;
```
brian-patrick-3, line 150: Why opacity on the icon?

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
```diff
@@ -135,11 +128,38 @@ const SourcesContent = () => html`
...
+
+    .separator {
+      width: 1px;
+      height: 16px;
+      background-color: var(--color-border, #ccc);
+      margin: 0 4px;
+    }
+    kyn-link {
+      word-break: break-word;
+      overflow-wrap: break-word;
```
brian-patrick-3, line 161: Do we want to allow word break? Maybe the whole card layout could be more flexible, instead of CSS gird we could use flexbox so they could stretch (flex-grow: 1) and wrap more naturally. Something to consider.

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
tulasigudibanda: Added opacity as Figma shows it.

brian-patrick-3: Hmm I'm not seeing any opacity in here: https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=32640-7254&m=dev

tulasigudibanda:
> > > 1,3 - done
> > > 2. Waiting for Natalia to confirm.
> > > 4. Waiting for Natalia to confirm on this: Are sources always kyn-cards or can they even be deeply nested structures ? Because 'Show more' logic assumes a uniform layout and countable items — which would break for deeply nested structures with how items are selected/hidden .
> > 
> > 
> > Shouldn't it be our designers (Robert) confirming these items?
> > Slotted sources should be kyn-cards yes, but if the application does anything out of the ordinary like using extra divs, creating a wrapper component, or putting in really long links for example, it could mess with the structure if we're still controlling the number of visible cards. Just something to consider, it probably will be fine as-is.
> 
> Previously Robert told me to confirm about earth icon with Natalia . So posted it in the group chat (with Robert and Chandan) . Sure. I will keep the existing logic in that case unless Natalia gets back with something different

@brian-patrick-3 ,did not understand this part : why putting really long links would interfere with show more logic ?

tulasigudibanda:
@brian-patrick-3 , Natalia wants it "word NUMBER word" pattern as 
Show 2 more instead of Show more (2)
She says app can translate it accordingly.

brian-patrick-3:
> @brian-patrick-3 ,did not understand this part : why putting really long links would interfere with show more logic ?

I think their intention was they want to only show "one row" of links with the ability to reveal more. That becomes a bit problematic when we can't predict how long/wide the links/cards will be. Hence the suggestion to maybe just let the application add/control it, and drop it from our responsibility.

brian-patrick-3:
> @brian-patrick-3 , Natalia wants it "word NUMBER word" pattern as Show 2 more instead of Show more (2) She says app can translate it accordingly.

They can inject it as a modified textString then, and we can leave it without the number by default.

tulasigudibanda:
> > @brian-patrick-3 ,did not understand this part : why putting really long links would interfere with show more logic ?
> 
> I think their intention was they want to only show "one row" of links with the ability to reveal more. That becomes a bit problematic when we can't predict how long/wide the links/cards will be. Hence the suggestion to maybe just let the application add/control it, and drop it from our responsibility.

ok. With long links , they wrap accordingly 

> > @brian-patrick-3 , Natalia wants it "word NUMBER word" pattern as Show 2 more instead of Show more (2) She says app can translate it accordingly.
> 
> They can inject it as a modified textString then, and we can leave it without the number by default.

ok. I will just leave it as 'Show more' by default

tulasigudibanda:
> > @brian-patrick-3 ,did not understand this part : why putting really long links would interfere with show more logic ?
> 
> I think their intention was they want to only show "one row" of links with the ability to reveal more. That becomes a bit problematic when we can't predict how long/wide the links/cards will be. Hence the suggestion to maybe just let the application add/control it, and drop it from our responsibility.

With long links it aligns like this : 

https://github.com/user-attachments/assets/e690f7d2-845b-44eb-9419-944b16822bd7

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
tulasigudibanda: I will try with flex-box instead of grid

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
tulasigudibanda: > Maybe use `gap` here instead so you don't have to add padding to individual flex items like this link and the globe icon.

Since gap only adds spacing between the flex items and not to the left of .source-icon and right of the link, I have added padding

brian-patrick-3: Right, but in that case you can just adjust the container/card padding. Technically fine either way just wanted to offer the idea.

tulasigudibanda:
On checking with Natalia , she is proposing some UX updates :
image

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
tulasigudibanda: I will remove it. Probably I highlighted the incorrect item in Figma

brian-patrick-3:
I think clickable cards with regular text is a fine idea. Really should be going through our designers though.

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
tulasigudibanda: Added gap + padding on container

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
```diff
@@ -135,11 +128,38 @@ const SourcesContent = () => html`
...
   <style>
-    .card-title div {
-      @include typography.type-ui-01;
-      color: var(--kd-color-text-level-primary);
-      font-size: 16px;
-      font-weight: 500;
+    kyn-card {
+      width: 100%;
+    }
+    kyn-card::part(card-wrapper) {
```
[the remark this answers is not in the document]
tulasigudibanda: @brian-patrick-3  , but ::slotted cannot target nested parts within those slotted elements, like the ::part() pseudo-element of a slotted kyn-card.

tulasigudibanda:
> > @tulasigudibanda
> > 
> > 1. Show more/show less button text color in dark mode is not visible on hover.
> > 2. Observed accessibility violations in dark mode
> 
> 'Show more' button text color is --kd-color-background-button-primary-ai-default in figma which is not the default. Hence the issue. Second issue is with the kyn-link inside the cards is throwing color contrast violation in dark mode I am waiting for Robert to check and get back on both of these.

1. Robert updated UX to default ghost button color
2. As per new requirement from AI team, kyn-link for source is changed to plain text inside the card . Instead the card is now made clickable

src/components/ai/sourcesFeedback/aiSourcesFeedback.stories.js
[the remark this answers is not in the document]
brian-patrick-3: That's a shame. In that case, we may need to add a prop to the card component to reduce the wrapper padding. The alternative would be to add the card-wrapper shadow part to our docs, advertise it's availability.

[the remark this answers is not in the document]
tulasigudibanda: > In that case, we may need to add a prop to the card component to reduce the wrapper padding. The alternative would be to add the card-wrapper shadow part to our docs, advertise it's availability.

Sure. Please let me know which method do you prefer?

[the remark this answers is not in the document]
brian-patrick-3: Let's just add the shadow part to the docs. Should be a comment something like `@part` at the top similar to how we show slots and events.

brian-patrick-3 (approved):

srpriyankashetty (approved):
Feedback Panel updates:
Verified that feedback textarea heigh is reduced to 3 rows
Verified that feedback panel allows textarea resize.

Source Panel Updates:
Verified that show more link is changed to in-line
Verified that padding and margin is reduced
Verified that count is removed from Show More button
Verified that links in source blocks are removed and changed to cards
Verified the icon and text are as per Figma
Verified that Source cards are changed to clickable cards
Verified the changes in both light and dark mode
Verified that source card boarder is matching with AI **(Figma must be updated)** 
Verified that on hover the color of the card is changed as per the Figma

PR #438, opened 2025-04-01
feat(pagination): add page selector dropdown

tulasigudibanda:
Added dropdown for page number selection in the Pagination component

[AB#2209974*](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2209974)

[AHA Request](https://kyndrylshidoka.ideas.aha.io/ideas/KSDESIGN-I-125)

### DIFF
--- a/src/components/reusable/pagination/pagination-navigation-buttons.ts
+++ b/src/components/reusable/pagination/pagination-navigation-buttons.ts
@@ -44,6 +44,12 @@ export class PaginationNavigationButtons extends LitElement {
   // Constant representing the smallest possible page number
   private readonly SMALLEST_PAGE_NUMBER = 1;
 
+  /** Label for the page size dropdown. Required for accessibility.
+   * @internal
+   */
+  @property({ type: String })
+  pageNumberLabel = 'Page number';
+
   /**
    * Handles the button click event, either moving to the next page or previous page
    * @param {boolean} next - If true, will move to the next page, otherwise to the previous page
@@ -99,8 +105,10 @@ export class PaginationNavigationButtons extends LitElement {
       <span class="page-range" role="status" aria-live="polite">
         <kyn-dropdown
           name="page-number"
+          label="${this.pageNumberLabel}"
+          ?hideLabel=${true}
+          inline
           size="sm"
-          placeholder="Page number"
           value="${this.pageNumber.toString()}"
           @on-change=${(e: CustomEvent) => this.handleChange(e)}
         >
--- a/src/components/reusable/pagination/pagination-navigation-buttons.ts
+++ b/src/components/reusable/pagination/pagination-navigation-buttons.ts
@@ -1,9 +1,10 @@
 import { unsafeSVG } from 'lit-html/directives/unsafe-svg.js';
 import { html, LitElement } from 'lit';
-import { property, customElement } from 'lit/decorators.js';
+import { property, customElement, state } from 'lit/decorators.js';
 
 // Import required components and icons
 import '../button';
+import '../dropdown';
 import chevLeftIcon from '@kyndryl-design-system/shidoka-icons/svg/monochrome/16/chevron-left.svg';
 import chevRightIcon from '@kyndryl-design-system/shidoka-icons/svg/monochrome/16/chevron-right.svg';
 
@@ -36,6 +37,10 @@ export class PaginationNavigationButtons extends LitElement {
   @property({ type: Object })
   textStrings: any = {};
 
+  /** Available options for the page number. */
+  @state()
+  pageNumberOptions: Array<number> = [];
+
   // Constant representing the smallest possible page number
   private readonly SMALLEST_PAGE_NUMBER = 1;
 
@@ -53,6 +58,22 @@ export class PaginationNavigationButtons extends LitElement {
         detail: { value: currentPage },
         bubbles: true, // Allows parent components to catch it
         composed: true, // Required for the event to pass through the Shadow DOM boundary
+      })
+    );
+  }
+
+  /**
+   * Handles the dropdown change event.
+   * @param {CustomEvent} event
+   */
+  private handleChange(event: CustomEvent) {
+    this.pageNumber = event.detail.value;
+
+    this.dispatchEvent(
+      new CustomEvent('on-page-number-change', {
+        detail: { value: Number(event.detail.value) },
+        bubbles: true,
+        composed: true,
       })
     );
   }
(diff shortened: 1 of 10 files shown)


src/components/reusable/pagination/pagination-navigation-buttons.ts
```diff
@@ -36,6 +37,10 @@ export class PaginationNavigationButtons extends LitElement {
   @property({ type: Object })
   textStrings: any = {};
 
+  /** Available options for the page number. */
+  @state()
+  pageNumberOptions: Array<number> = [];
```
brian-patrick-3, line 42: This looks like a private property? Mark it with an `@internal` comment, and prefix with an `_`

src/components/reusable/pagination/pagination-navigation-buttons.ts
```diff
@@ -76,8 +97,22 @@ export class PaginationNavigationButtons extends LitElement {
       </kyn-button>
 
       <span class="page-range" role="status" aria-live="polite">
-        ${this.pageNumber} ${this.textStrings.of} ${this.numberOfPages}
-        ${this.textStrings.pages}
+        <kyn-dropdown
```
brian-patrick-3, line 100: 1. This dropdown needs a label, it can be hidden visually with `hideLabel`. This will also fix the alignment.
2. Use the `inline` prop so it matches the items per page dropdown.

brian-patrick-3 (changes requested):

src/components/reusable/pagination/pagination-navigation-buttons.ts
```diff
@@ -76,8 +97,22 @@ export class PaginationNavigationButtons extends LitElement {
...
+          name="page-number"
+          size="sm"
+          placeholder="Page number"
```
brian-patrick-3, line 103: Can probably remove this placeholder as it will never be visible, that way we won't need to include it in textStrings for customization/localization either.

src/components/reusable/pagination/pagination-navigation-buttons.ts
```diff
@@ -93,6 +128,15 @@ export class PaginationNavigationButtons extends LitElement {
       </kyn-button>
     `;
   }
+
+  override willUpdate(changedProps: any) {
+    if (changedProps.has('numberOfPages')) {
```
brian-patrick-3, line 133: You may also need to also update this when changing other props like `count` and `pageSize`

tulasigudibanda: When pageSize is changed in UI, numofPages is getting updated currently. Am I missing anything?

brian-patrick-3: I think I was seeing that the current page isn't resetting. Ex current page: 7, change page size to 50, shows page 7 of 2

tulasigudibanda: I see the page number getting reset :
https://github.com/user-attachments/assets/573c4db2-0b38-4264-9b0b-6e5da928433c

brian-patrick-3: I see, the example story table handles that. The pagination does not by itself.

tulasigudibanda: Reset is happening in the pagination component too : https://deploy-preview-438--shidoka-applications.netlify.app/?path=/story/components-pagination--pagination .
Please me know how you tested it .

brian-patrick-3:
What happened to the text wrapping?
Screenshot 2025-04-02 134905

tulasigudibanda:
> Seems to be Firefox only. Maybe due to the new `display: flex;` CSS?

Works fine in forefox(136.0.4 (aarch64))  too for me.

brian-patrick-3 (approved):

srpriyankashetty (approved):
Verified in Firefox and edge browser, looks fine.
Verified the dropdown value with different items per page
Verified back and forth navigation with arrow and looks fine
Verified the hide navigation button.
Verified in both light and dark mode

Thought: Can we have option to hide the page selector dropdown alone.! not the complete navigator.
I don't see a control, specific to dropdown alone. If it's doable form backend then fine. @brian-patrick-3, @tulasigudibanda 

image

tulasigudibanda:
> Verified in Firefox and edge browser, looks fine. Verified the dropdown value with different items per page Verified back and forth navigation with arrow and looks fine Verified the hide navigation button. Verified in both light and dark mode
> 
> Thought: Can we have option to hide the page selector dropdown alone.! not the complete navigator. I don't see a control, specific to dropdown alone. If it's doable form backend then fine. @brian-patrick-3, @tulasigudibanda

@srpriyankashetty , we don't have for hide option for page selector dropdown alone. Could you please clarify the use case you're trying to address with this :  Do you mean to switch between dropdown and simple page number rendering ?
image

srpriyankashetty:
> > Verified in Firefox and edge browser, looks fine. Verified the dropdown value with different items per page Verified back and forth navigation with arrow and looks fine Verified the hide navigation button. Verified in both light and dark mode
> > Thought: Can we have option to hide the page selector dropdown alone.! not the complete navigator. I don't see a control, specific to dropdown alone. If it's doable form backend then fine. @brian-patrick-3, @tulasigudibanda
> 
> @srpriyankashetty , we don't have for hide option for page selector dropdown alone. Could you please clarify the use case you're trying to address with this : Do you mean to switch between dropdown and simple page number rendering like in current master ? image

srpriyankashetty:
yes @tulasigudibanda, just wanted to know if it's doable to hide the dropdown alone if not required.!
If in case anyone comes back, just to be sure.!
CC: @brian-patrick-3

tulasigudibanda:
> yes @tulasigudibanda, just wanted to know if it's doable to hide the dropdown alone if not required.! If in case anyone comes back, just to be sure.! CC: @brian-patrick-3

Users can still navigate using the left and right arrows if they prefer not to select from the page number dropdown. In a narrower screen view (such as on mobile), only the left and right navigation buttons are displayed.
There’s no issue in displaying available details when there is sufficient screen space. Additionally, the hideNavigationButtons option is already available, which hides the entire page navigation section. 
It seems unlikely that the specific requirement of hiding only the dropdown will come up again, but we can revisit it if the need arises.

wbarbee (approved):

PR #439, opened 2025-04-02
feat(SliderInput): added slider input component

kumarivijeta:
## Summary

A Slider input is an input element that allows users to select a number from a specified range using a slider control.

## ADO Story

Feature [AB#2211107](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2211107)
Fixes [AB#2241662](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2241662)
Fixes [AB#2242722](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2242722)

## Figma Link

[Link here](https://www.figma.com/design/rC5XdRnXVbDmu3vPN8tJ4q/2.1-Edinburgh?node-id=4755-2877&p=f&m=dev)

## Notes

- Detailed change notes
- can go here

## Checklist

  - [ x] Ran the `analyze` command to update Storybook docs.

## Testing Instructions


src/components/reusable/sliderInput/sliderInput.ts
```diff
@@ -0,0 +1,296 @@
...
+            aria-invalid=${this._isInvalid}
+            aria-describedby=${this._isInvalid ? 'error' : ''}
+            style=${Object.entries(styles)
+              .map(([key, value]) => `${key}: ${value}`)
+              .join(';')}
+            @input=${(e: any) => this.handleInput(e)}
+          />
+          <span>${this.max}</span>
+
+          <input
```
brian-patrick-3, line 130: I don't think we need to include a number input inside the range.

src/components/reusable/sliderInput/sliderInput.ts
```diff
@@ -0,0 +1,296 @@
...
+                  ${this.invalidText || this._internalValidationMsg}
+                </div>
+              `
+            : null}
+        </div>
+      </div>
+    `;
+  }
+
+  private fillTrackBackground() {
```
brian-patrick-3, line 164: Can the track/progress just be styled with CSS? 
Ref: https://blog.logrocket.com/creating-custom-css-range-slider-javascript-upgrades/

src/components/reusable/sliderInput/sliderInput.ts
```diff
@@ -0,0 +1,296 @@
...
+      validity = { ...validity, rangeOverflow: true };
+      this._internalValidationMsg = `Value must be less than or equal to ${max}`;
+      this._inputEl.setCustomValidity(
+        `Value must be less than or equal to ${max}`
+      );
+    } else {
+      this._isInvalid = false;
+      this._internalValidationMsg = '';
+      this._inputEl.setCustomValidity('');
+    }
```
brian-patrick-3, line 250: These min/max validations should all come natively.

src/components/reusable/sliderInput/sliderInput.ts
```diff
@@ -0,0 +1,296 @@
...
+      //
+      inputEl.style.background = `linear-gradient(to right, ${getTokenThemeVal(
+        '--kd-color-background-button-primary-state-default'
+      )} ${percentage}%, #dee1e2 ${percentage}%, #dee1e2 100%)`;
+    }
+  }
+
+  private handleInput(e: any) {
+    if (this._isInvalid) return;
+    this.value = this.shadowRoot?.querySelector('input')?.value;
```
brian-patrick-3, line 180: You should be able to get the value from the event data without having to query the element.

src/components/reusable/sliderInput/sliderInput.ts
```diff
@@ -0,0 +1,296 @@
...
+      this._internals.reportValidity();
+    }
+  }
+
+  override updated(changedProps: any) {
+    // preserve FormMixin updated function
+    this.value = this.shadowRoot?.querySelector('input')?.value;
+    if (this.value) {
+      this.fillTrackBackground();
+    }
```
brian-patrick-3, line 277: This is running on every update.

brian-patrick-3 (commented):
Just a few notes after a quick glace

src/components/reusable/sliderInput/sliderInput.ts
kumarivijeta: To give more flexibility on user input. I have added this one. Additionally I have added a control to hide/show this input box.

brian-patrick-3: I think if apps really want it, they can have a separate number input and link them. Hesitant to bake it in because the MDN page does not show any native examples of that.

src/components/reusable/sliderInput/sliderInput.ts
kumarivijeta: Addressed.

src/components/reusable/sliderInput/sliderInput.ts
kumarivijeta: I tried to update the progress track using CSS but it doesn't seem looking good .

In the image if you notice  1st example shows with using css and the rest is using js.

image

src/components/reusable/sliderInput/sliderInput.ts
kumarivijeta: If `this.fillTrackBackground()` called on `handleInput` & `handleKeydown`

 Following use case fails was failing 
 
1. On load, when the value is predefined, the background color of the track does not update.

Screenshot 2025-04-04 at 4 49 01 PM


2. When the user scrolls the slider, the track background color fills correctly, but if the value changes, the track progress color does not update properly.
Screenshot 2025-04-04 at 4 49 59 PM

Since the slider progress track background is being updated logically, and the value is continuously updating during the slide, the `update` method would be the ideal place to address the above use case. Also it ensures that the track progress color is properly updated whenever the value changes.

brian-patrick-3 (changes requested):
1. Can you use the native `<datalist>` method of adding ticks and labels? https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range#adding_labels This should simply things greatly.
2. It seems unnecessary to me to have both ticks, and dots/segments on the line. I would personally suggest getting rid of the dots and just use native tick marks.
3. The slider handle doesn't seem to align with the tickmarks always, particularly the first and last ticks. It should always be centered over the ticks.
4. Is there an option to show ticks both with and without the handle snapping to each tick?
6. In the number input, maybe we should hide the arrows to match UX unless they like having them.
7. Looking at the UX, there should be some sort of border/outline on the handle, or a gap, to visually separate it from the slider track.
8. The tooltip is a nice touch for when the number input is hidden!


> 4. Is there an option to show ticks both with and without the handle snapping to each tick? 
    The ticks has been calculated based on step, min and max values that's why handle is snapped to each tick.
> 6. Looking at the UX, there should be some sort of border/outline on the handle, or a gap, to visually separate it from the slider track.
On focus added a outline on handle but due to webkit box shadow handle outline is not getting applied specially on Chrome and Safari.  But it is works for firefox.

src/components/reusable/sliderInput/sliderInput.ts
```diff
@@ -0,0 +1,505 @@
...
+const _defaultTextStrings = {
+  error: 'Error',
+};
+
+/**
+ * Slider Input.
+ * @fires on-input - Captures the input event and emits the selected value and original event details.
+ * @prop {number} min - The minimum value.
+ * @prop {number} max - The maximum value.
+ * @prop {number} step - A value determining how much the value should increase/decrease by moving the thumb by mouse. If a value other than 1 is provided and the input is not hidden, the new step requirement should be added to a visible label. Values outside the step increment will be considered invalid..
```
brian-patrick-3, line 24: You don't need to document props at the top here, just do it inline like the others.

src/components/reusable/sliderInput/sliderInput.ts
```diff
@@ -0,0 +1,505 @@
...
+      );
+
+      // Calculate the current thumb position as a percentage
+      const thumbPercentage = ((clampedValue - min) / (max - min)) * 100;
+
+      // Compare the tick's position with the thumb's position
+      if (tickPercentage <= thumbPercentage) {
+        // Turn the tick white if the thumb has passed it
+        tick.style.backgroundColor = getTokenThemeVal(
+          '--kd-color-background-accent-subtle'
```
brian-patrick-3, line 404: This seems unnecessary, the tick is already styled with a color token in the CSS. Let's remove the mutation observer entirely.

kumarivijeta:
> 1. I still see some slight misalignment of scale tick marks with the handle.
> 2. In firefox, the ticks on the track are slightly larger than the track.

> 1. I still see some slight misalignment of scale tick marks with the handle.

> 1. I still see some slight misalignment of scale tick marks with the handle.

referring the below ?

image

> 2. In firefox, the ticks on the track are slightly larger than the track.

Somehow, 1px extra is getting added on Firefox . Will check this out.

brian-patrick-3 (commented):
Can you please also add a slider to the Forms page? Need to test it out in a form context to.

brian-patrick-3 (commented):
It seems like the scale is still dependent on step. We need to be able to show the scale without it being tied to step, ex: continuous story with visible scale.

kumarivijeta:
> It seems like the scale is still dependent on step. We need to be able to show the scale without it being tied to step, ex: continuous story with visible scale.

  Yes, Scale dependent on min, max and step values. If no min,max and step value provided it takes default value 0,100 and 1 and Due to smaller width of slider continuous varient with scale will not be appropriate.  I think `Discrete varient with scale` would be more appropriate place 
  
  
slider continuous varient 

image

brian-patrick-3:
> > It seems like the scale is still dependent on step. We need to be able to show the scale without it being tied to step, ex: continuous story with visible scale.
> 
> Yes, Scale dependent on min, max and step values. If no min,max and step value provided it takes default value 0,100 and 1 and Due to smaller width of slider continuous varient with scale will not be appropriate. I think `Discrete varient with scale` would be more appropriate place
> 
> slider continuous varient
> image

Right this screenshot is exactly why we need to be able to specify the tick marks separately from the `step`

brian-patrick-3:
Custom Labels idea is close, but we still need scale marks without labels as well based on figma.
Screenshot 2025-04-21 090919

kumarivijeta:
> Custom Labels idea is close, but we still need scale marks without labels as well based on figma. Screenshot 2025-04-21 090919


To get the scale mark w/o labels , all we have to do pass empty string in custom label .

image

image

wbarbee (approved):

srpriyankashetty (approved):
1. Verified the slider component in both light and dark mode and looks fine
2. Verified by setting minimum and maximum value
3. Verified changing the step to multiple values and looks fine
4. Verified the editable input
5. Verified the disable
6. Verified the scale marker
7. Verified the slider with tick mark pattern
8. Verified the other props like invalid text, caption, hide label and looks fine
9. Verified the sider icon on click and default
10. Verified the track filled color in both light and dark mode
11. Verified the number input not accepting negative values
12. Verified that number input accepts value based on the step number and error message is displayed if not met.

PR #449, opened 2025-04-17
feat(date-time pickers): optional programmatic clear method added to time, date, date-range pickers

wbarbee:
## Summary

Dev was having issues implementing programmatic clearing from a tag that was external to the date-range picker component. This PR adds in an _optional_ programmatic clear (Lit @method) to make this easier to achieve across all three flatpickr-related components, enhancing the reliability of the programmatic clearing feature.

1. Improved error handling

2. Flag management

3. Consistency across components

## Checklist

## Testing Instructions

Added new Programmatic Clear Variant Examples to each of the three components to illustrate how this can be achieved from an external button.

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -5360,6 +5360,25 @@
             },
             {
               "kind": "method",
+              "name": "clear",
+              "privacy": "public"
+            },
+            {
+              "kind": "method",
+              "name": "_clearInput",
+              "privacy": "private",
+              "parameters": [
+                {
+                  "name": "options",
+                  "default": "{ reinitFlatpickr: true }",
+                  "type": {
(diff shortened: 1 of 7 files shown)


brian-patrick-3:
Hmm I'm not a huge fan of adding methods that have to be called by querying the element first, it's kind of a "last resort" solution. Is there a reason these components can't be cleared properly just by updating the `value`?

wbarbee:
> Hmm I'm not a huge fan of adding methods that have to be called by querying the element first, it's kind of a "last resort" solution. Is there a reason these components can't be cleared properly just by updating the `value`?

i just checked with the dev who was having issues. he was able to get it working yesterday with the existing logic. just wasn't initializing the value correctly.

i'll keep this branch available locally to return to, in case more people are having issues in the future

PR #460, opened 2025-04-29
feat: add new button enhancement to dropdown

tulasigudibanda:
## Summary

Dropdown Enhancements:

Support for adding a new option via an input and button at the top of the options menu (emits an event, but does not update the UI directly).
Support for removing individual options using an "X" button on the right side (emits an event, but does not remove the option from the UI).

[AB#2241842](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2241842)

## Figma Link

[Figma](https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=33866-47239&p=f&m=dev)

## To Do
Working on accessibility
Fixing existing accessibility bugs :
[BUG 2259898](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2259898)


kumarivijeta:
@tulasigudibanda 
Couple of Observation:
1. Right side checkmark color is not matching with figma.
2. X & [x]  is not aligned.

src/components/reusable/dropdown/dropdownOption.ts
```diff
@@ -3,14 +3,17 @@ import { LitElement, html } from 'lit';
...
 
 import checkIcon from '@kyndryl-design-system/shidoka-icons/svg/monochrome/16/check.svg';
+import clearIcon from '@kyndryl-design-system/shidoka-icons/svg/monochrome/16/close-simple.svg';
 
 import DropdownOptionScss from './dropdownOption.scss';
 
 /**
  * Dropdown option.
  * @fires on-click - Emits the option details to the parent dropdown.
+ * @fires on-remove - Emits the option that is removed.
```
brian-patrick-3, line 16: Seems like we're emitting this event from both the option and the parent dropdown. I would go with one or the other, not both.

src/components/reusable/dropdown/dropdown.ts
```diff
@@ -337,6 +349,26 @@ export class Dropdown extends FormMixin(LitElement) {
               @keydown=${(e: any) => this.handleListKeydown(e)}
               @blur=${(e: any) => this.handleListBlur(e)}
             >
+              <!-- Add New Option -->
```
brian-patrick-3, line 352: I think the "add new option" feature needs to be off/hidden by default.

brian-patrick-3: These elements also do not seem to be keyboard interactive. I'm not quite sure what the best way to do that would be.

brian-patrick-3 (changes requested):
Maybe we should add another "data driven options" story that shows a functional example of adding an option.

src/components/reusable/dropdown/dropdownOption.ts
```diff
@@ -95,11 +98,35 @@ export class DropdownOption extends LitElement {
 
         ${this.selected && !this.multiple
           ? html`<span class="check-icon">${unsafeSVG(checkIcon)}</span>`
-          : null}
+          : html`<kyn-button
```
brian-patrick-3, line 101: These are also not keyboard interactive.

src/components/reusable/dropdown/dropdown.ts
```diff
@@ -337,6 +349,26 @@ export class Dropdown extends FormMixin(LitElement) {
...
+                  class="add-option-input"
+                  type="text"
+                  placeholder=${this._textStrings.addItem}
+                  .value=${this.newOptionValue}
+                  @input=${(e: any) => this._handleInputNewOption(e)}
+                  @mousedown=${(e: Event) => e.stopPropagation()}
+                  @keydown=${(e: KeyboardEvent) => e.stopPropagation()}
+                  @focus=${(e: KeyboardEvent) => e.stopPropagation()}
+                />
+                <kyn-button
```
brian-patrick-3, line 364: This button appears to be escaping it's container to the right side, creating a horizontal scrollbar in the menu.

src/components/reusable/dropdown/dropdownOption.ts
tulasigudibanda: > Seems like we're emitting this event from both the option and the parent dropdown. I would go with one or the other, not both.

Is it okay to dispatch the on-remove-option event from kyn-dropdown-option and include it in stories.js like this: @on-remove-option=${(e) => action(e.type)(e)} to demonstrate that the event can be hooked? The event fires correctly, but it doesn't appear in the Storybook Events panel.

brian-patrick-3: I'm not following. You have `@on-remove` coming from the option, and `@on-remove-option` from the parent dropdown. I see both appearing in the storybook props tables. It would be confusing for devs to have both of those event hooks available. I would probably just emit it directly from the option instead of bubbling to the parent and re-emitting it.

tulasigudibanda: Removed parent event and updated @on-remove coming from the option to @on-remove-option

src/components/reusable/dropdown/dropdown.ts
tulasigudibanda: I have added the .dropdown min-width to 20rem to take care of this . It looks like this for me, even in mobile view :
image

brian-patrick-3: That's probably going to be too large for 320px screen size. With 16px page gutters, the max it can be is 288px

tulasigudibanda: ok. So do we need to put max-width on .add-option-input, so 'Add' button won't escape to the right?

brian-patrick-3: I think you can just add `min-width: 0` to the input to fix it

brian-patrick-3: kd-btn also has a `min-width: 96px` in code for some reason making the button look larger here. wonder if we should remove that

tulasigudibanda: It becomes so small 
image

tulasigudibanda: Yes. It looks good if we remove that 96px.

brian-patrick-3: That small, maybe allowing flex wrap would help too. Tough at that size

tulasigudibanda: With flex-wrap: wrap, input and button always wraps unless we set a min-width on the dropdown.

brian-patrick-3 (approved):

wbarbee (approved):

srpriyankashetty:
@tulasigudibanda, please check the following
1. Add button UI is not matching with Figma added screenshot (edge and chrome)
2. Make few options selection, delete any one newly added option. Except Option 2, all other options are deselected
Attached video
image

https://github.com/user-attachments/assets/42791675-03ca-4b7a-a518-da742fd42724

src/components/reusable/dropdown/dropdown.ts
```diff
@@ -322,33 +351,34 @@ export class Dropdown extends FormMixin(LitElement) {
...
+                        class="add-option-input"
+                        type="text"
+                        placeholder=${this._textStrings.addItem}
+                        .value=${this.newOptionValue}
+                        aria-label="Add new option"
+                        @input=${this._handleInputNewOption}
+                        @keydown=${this._onAddOptionInputKeydown}
+                        @focus=${this._onAddOptionInputFocus}
+                      />
+                      <button
```
brian-patrick-3, line 374: Good catch on the button, yeah this should be a kyn-button

brian-patrick-3 (changes requested):

brian-patrick-3 (approved):

brian-patrick-3:
@srpriyankashetty 

>     1. Add button UI is not matching with Figma added screenshot (edge and chrome)
>     2. Make few options selection, delete any one newly added option. Except Option 2, all other options are deselected

I believe both are resolved now.

kumarivijeta (approved):

srpriyankashetty:
@brian-patrick-3  and @kumarivijeta 
Observations:
1. When tried to add option from Signle select, the options are not getting add in the dropdown list
2. Tooltip for delete the created option (x mark) - confirm if tooltip is required.

image

srpriyankashetty (approved):
Verified that add button is added as per Figma
Verified that able to add options using Add button
Verified that able to add options with special characters and alpha numeric
Verified that created options are listing in sorted way
Verified that selected options are showing in tags
Verified that selected options are in sorted pattern
Verified that on deleting the created option, the dropdown list is getting updated
Verified that deleted option is also removed form the selected tags if any
Tooltip is displayed for deleting the added option (x mark)

srpriyankashetty:
@brian-patrick-3, Please confirm on the following
1. Add button in the dark mode has no boarders or background
2. There are some accessibility violations showing on adding the option

image

brian-patrick-3:
>     2. There are some accessibility violations showing on adding the option (Note: dropdown overlay must be open)
I made some changes, I don't think it's possible fix the "interactive controls must not be nested" error with the current design. We'll either have to live with it, or go back to the drawing board.

PR #466, opened 2025-05-14
fix(dropdown): selections and disabled options

brian-patrick-3:
## Summary

A handful of fixes to work around selection/deselection when disabled and disabled + selected options are involved.

1. Clear All tag will not remove disabled selected options
2. Select All option somehow went missing, and I've restored that and excluded disabled options from it's behavior.
3. Tags for disabled options will now be disabled also.

### DIFF
--- a/src/components/reusable/dropdown/dropdown.stories.js
+++ b/src/components/reusable/dropdown/dropdown.stories.js
@@ -147,7 +147,7 @@ export const SingleSearchable = {
 };
 
 export const MultiSelect = {
-  args: { ...args, value: ['1'] },
+  args: { ...args, value: ['1', '3'] },
   render: (args) => {
     return html`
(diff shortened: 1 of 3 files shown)


wbarbee (approved):
- [x] Clear All tag will not remove disabled selected options
- [x] Select All option somehow went missing, and I've restored that and excluded disabled options from it's behavior.
- [x] Tags for disabled options will now be disabled also.

kumarivijeta:
- [x] Clear All tag will not remove disabled selected options
- [x] Select All option somehow went missing, and I've restored that and excluded disabled options from it's behavior.
- [x] Tags for disabled options will now be disabled also.

One obervation:
1. For add new option varient , Add new item and hover on recently added option item doesn't show correct text on tooltip. 

image

PR #509, opened 2025-06-25
fix(WidgetSideDrawer): added admin tab

kumarivijeta:
## Summary

Added Admin Tab

## ADO Story 

[AB#2097838](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2097838)

## Figma Link

[Figma](https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=36270-17768&m=dev)

## Dependancy PR

https://github.com/kyndryl-design-system/shidoka-applications/pull/508  Selected Widget State

## Checklist

### DIFF
--- a/src/components/global/header/headerCategory.scss
+++ b/src/components/global/header/headerCategory.scss
@@ -12,6 +12,7 @@
 }
 
 .heading {
+  @include typography.type-ui-03;
   display: flex;
   align-items: center;
   gap: 8px;
@@ -19,12 +20,10 @@
   font-weight: 500;
   text-transform: uppercase;
   padding: 20px 12px 0 12px;
-  // border-bottom: 1px solid var(--kd-color-border-level-tertiary);
 
   &.left-padding {
     padding-left: 36px;
   }
-  @include typography.type-ui-03;
 }
 
 slot[name='icon']::slotted(*) {
--- a/src/components/global/header/headerFlyout.scss
+++ b/src/components/global/header/headerFlyout.scss
@@ -39,11 +39,11 @@
 }
 
 .menu-label {
+  @include typography.type-ui-03;
   color: var(--kd-color-text-title-secondary);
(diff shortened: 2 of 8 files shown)


src/components/reusable/button/button.ts
```diff
@@ -202,7 +202,7 @@ export class Button extends LitElement {
               formmethod=${ifDefined(this.formmethod)}
               @click=${(e: Event) => this.handleClick(e)}
             >
-              <span>
+              <span part="button-content">
```
brian-patrick-3, line 205: There is a problem with this approach. Since we don't expose `part`s to developers, they won't be able to achieve this button design with split text and icon. It needs to be built-in with a prop or something to change the button from centered content to split content (`justify-content: space-between`).

brian-patrick-3: Right, but my point is that external devs won't have access to the CSS part to re-create this.

kumarivijeta: Added a fullWidth prop in the btn component to achieve the intended  UX.

src/components/reusable/button/button.ts
```diff
@@ -148,6 +148,22 @@ export class Button extends LitElement {
   @query('.button')
   accessor _btnEl!: any;
 
+  /** Aligns button text and icon at the edges when `kyn-button` has a set width.*/
+  @property({ type: Boolean })
+  accessor fullWidth = false;
```
brian-patrick-3, line 153: I find `fullWidth` a bit confusing for what this does. Maybe call it `splitLayout` instead? Otherwise, this seems like a good approach now.

kumarivijeta: Renamed fullWidth -> fluidLayout prop.

PR #530, opened 2025-07-08
feat(textinput): autocomplete enable/disable

kumarivijeta:
## Summary

Added autocomplete property to turn off/on native browser suggestions for form input

1. ColorInput
2. TextInput

Added Autocomplete off  to the <input>

1. Search
2. MultiInput Field
3. DatePicker 
4. Timepicker

## ADO Story Link

[AB#2360123](https://dev.azure.com/Kyndryl/Shidoka%20-%20Bridge%20Design%20System/_workitems/edit/2360123/)

## Figma Link

NA

## Notes

- Detailed change notes
- can go here

## Checklist

## Screenshots

**1. Text Input**
**When Autocomplete 'on'**

image

**When Autocomplete 'off'**

image

**3.Color Input** 

**When Autocomplete 'on'**

image

**When Autocomplete 'off'**

image

### DIFF
--- a/src/components/reusable/colorInput/colorInput.ts
+++ b/src/components/reusable/colorInput/colorInput.ts
@@ -47,6 +47,10 @@ export class ColorInput extends FormMixin(LitElement) {
   @property({ type: Object })
   accessor textStrings = _defaultTextStrings;
 
+  /** AutoComplete state */
+  @property({ type: Boolean })
+  accessor autoComplete = false;
+
   /** Internal text strings.
    * @internal
    */
@@ -114,6 +118,7 @@ export class ColorInput extends FormMixin(LitElement) {
             aria-label=${this._textStrings.colorTextInput}
             aria-invalid=${this._isInvalid}
             aria-describedby=${this._isInvalid ? 'error' : ''}
+            autocomplete=${this.autoComplete ? 'on' : 'off'}
             @input=${(e: any) => this.handleTextInput(e)}
           />
           <div class="caption-error-count">
--- a/src/components/reusable/colorInput/colorInput.ts
+++ b/src/components/reusable/colorInput/colorInput.ts
@@ -47,9 +47,9 @@ export class ColorInput extends FormMixin(LitElement) {
   @property({ type: Object })
   accessor textStrings = _defaultTextStrings;
 
-  /** AutoComplete state */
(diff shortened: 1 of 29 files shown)


src/components/reusable/multiInputField/multiInputField.ts
```diff
@@ -91,6 +91,10 @@ export class MultiInputField extends FormMixin(LitElement) {
   @property({ type: String })
   accessor pattern: string | undefined = undefined;
 
+  /** Toggles native browser autocomplete suggestions */
+  @property({ type: Boolean })
```
wbarbee, line 95: for multiInput, i already have `autoSuggestionDisabled ` (named as such because we wanted them active by default). i think this is essentially accomplishing the same thing as `autoComplete` here. it's early enough in the 'life' of this component that I don't think it would be too late to rename this property, but we shouldn't have both. 

https://github.com/kyndryl-design-system/shidoka-applications/blob/a2ade9b3a4f85deec43600b31fd81156dcd28546/src/components/reusable/multiInputField/multiInputField.ts#L66-L68

brian-patrick-3 (changes requested):
1. The `autocomplete` property should be a string that allows devs to pass any value/token supported by [the native spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete).
2. We should also include `textarea`.
3. What about date/time pickers?

src/components/reusable/multiInputField/multiInputField.ts
kumarivijeta: `autocomplete` prop  used to disable native browser suggestion. Observe on screenshot there are two suggestion panel shows one is browser owned suggestions and other is component owned suggestions.

image

wbarbee (approved):

src/components/reusable/colorInput/colorInput.ts
```diff
@@ -47,6 +47,10 @@ export class ColorInput extends FormMixin(LitElement) {
   @property({ type: Object })
   accessor textStrings = _defaultTextStrings;
 
+  /** AutoComplete state */
+  @property({ type: Boolean })
+  accessor autoComplete = false;
```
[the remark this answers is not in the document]
kumarivijeta: Fixed

PR #548, opened 2025-07-25
fix(date pickers): better dev width control

wbarbee:
## Summary

User alerted us in chat about inability/difficulty in controlling the width of the datepicker/daterangepicker/timepicker inputs. These modifications give the user more control over width, while setting a default width to not break any existing implementations of these components.

## Checklist

### DIFF
--- a/src/components/reusable/datePicker/datepicker.scss
+++ b/src/components/reusable/datePicker/datepicker.scss
@@ -4,18 +4,26 @@
 
 @include mixins.picker-base;
 
+:host {
+  display: inline-block;
+  width: 100%;
+  max-width: var(--kyn-date-range-picker-max-width, 250px);
+}
+
 input {
   padding: 0 50px 0 16px;
+}
+
+.input-wrapper,
+.flatpickr-wrapper,
+input {
+  width: 100%;
+  box-sizing: border-box;
 }
 
 .input-wrapper {
   position: relative;
   display: inline-block;
-  width: 100%;
-
-  .flatpickr-wrapper {
-    width: 100%;
-  }
 
   .input-icon,
   .clear-icon {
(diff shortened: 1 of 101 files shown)


src/components/reusable/datePicker/datepicker.scss
```diff
@@ -4,18 +4,26 @@
 
 @include mixins.picker-base;
 
+:host {
+  display: inline-block;
+  width: 100%;
+  max-width: var(--kyn-date-range-picker-max-width, 250px);
```
brian-patrick-3, line 10: We want `display: inline-block;` and `width: auto;` to be the defaults for all these form input components. `max-width` should probably just be `100%`.

src/components/reusable/daterangepicker/daterangepicker.scss
```diff
@@ -4,10 +4,23 @@
...
+
 input {
   padding: 0 50px 0 16px;
 }
 
+.input-wrapper,
+.flatpickr-wrapper,
+input {
+  width: 100%;
+  box-sizing: border-box;
```
brian-patrick-3, line 21: `box-sizing: border-box` should be included by default for all elements via the global.scss file.

src/components/reusable/textArea/textArea.ts
```diff
@@ -211,6 +211,10 @@ ${this.value}</textarea
     const event = new CustomEvent('on-input', {
       detail: {
         value: e.target.value,
+        isValid: !this._isInvalid,
+        validity: this.textareaEl.validity,
```
brian-patrick-3, line 215: Validation state can already be accessed externally by calling `checkValidity()` or `reportValidity()` on the element or the form.

src/components/reusable/datePicker/datepicker.scss
brian-patrick-3: `max-width: 100%` wouldn't change the width, just prevent it from overflowing the container.

src/components/reusable/textArea/textArea.ts
wbarbee: alright, reverted

PR #551, opened 2025-07-29
feat(enhanced dropdown): new dropdown child component

wbarbee:
## Summary

Optional `enhancedDropdownOption` child component that can be 'swapped' in/out for current default `dropdownOption` based on boolean value `enhanced` (false by default).

[AB#2397936](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/2397936)

## Figma Link

[Figma](https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library-2.0?node-id=13361-476952&m=dev)

## Checklist

## Screenshots
Screenshot 2025-07-30 at 9 18 05 AM
<br/>
Screenshot 2025-07-30 at 9 18 11 AM

### DIFF
--- a/src/components/reusable/dropdown/dropdown.ts
+++ b/src/components/reusable/dropdown/dropdown.ts
@@ -1078,14 +1078,6 @@ export class Dropdown extends FormMixin(LitElement) {
     }
   }
 
-  private _handleLabelKeyDown(e: KeyboardEvent): void {
-    if (this.disabled) return;
-    if (e.key === 'Enter' || e.key === ' ') {
-      e.preventDefault();
-      this._handleLabelClick();
-    }
-  }
-
   override connectedCallback() {
     super.connectedCallback();
 
--- a/src/components/reusable/dropdown/dropdown.ts
+++ b/src/components/reusable/dropdown/dropdown.ts
@@ -1,5 +1,5 @@
 import { unsafeSVG } from 'lit-html/directives/unsafe-svg.js';
-import { LitElement, html, unsafeCSS } from 'lit';
+import { LitElement, PropertyValues, html, unsafeCSS } from 'lit';
 import { customElement, property, state, query } from 'lit/decorators.js';
 import { classMap } from 'lit/directives/class-map.js';
 import DropdownScss from './dropdown.scss?inline';
@@ -73,7 +73,7 @@ export class Dropdown extends FormMixin(LitElement) {
   accessor placeholder = '';
 
   /** Listbox/drawer open state. */
-  @property({ type: Boolean })
+  @property({ type: Boolean, reflect: true })
   accessor open = false;
 
   /** Makes the dropdown searchable. */
@@ -123,6 +123,10 @@ export class Dropdown extends FormMixin(LitElement) {
   /** Controls direction that dropdown opens. */
   @property({ type: String })
   accessor openDirection: 'auto' | 'up' | 'down' = 'auto';
+
+  /** Controls the dropdown anchor type. */
+  @property({ type: String })
+  accessor dropdownAnchor: 'input' | 'button' = 'input';
 
   /** Is "Select All" box checked.
    * @internal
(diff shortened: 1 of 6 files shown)


src/components/reusable/dropdown/dropdown.ts
```diff
@@ -75,13 +73,17 @@ export class Dropdown extends FormMixin(LitElement) {
...
+  @property({ type: Boolean, reflect: true })
   accessor open = false;
 
   /** Makes the dropdown searchable. */
   @property({ type: Boolean })
   accessor searchable = false;
 
+  /** Makes the dropdown enhanced. */
+  @property({ type: Boolean })
+  accessor enhanced = false;
```
brian-patrick-3, line 85: I wonder if this could be automatically inferred based on child elements to make it slightly easier to use?

digutierrezberrocal (approved):

src/components/reusable/dropdown/dropdown.ts
```diff
@@ -122,6 +124,10 @@ export class Dropdown extends FormMixin(LitElement) {
   @property({ type: String })
   accessor openDirection: 'auto' | 'up' | 'down' = 'auto';
 
+  /** Controls the dropdown anchor type. */
+  @property({ type: String })
+  accessor dropdownAnchor: 'input' | 'button' = 'input';
```
brian-patrick-3, line 129: Is this also relevant for the non-enhanced dropdown? Might want to clarify.

src/components/reusable/dropdown/dropdown.ts
```diff
@@ -284,82 +299,93 @@ export class Dropdown extends FormMixin(LitElement) {
...
-                        : this.text}
-                    </span>
-                  `}
-
-              <span class="arrow-icon">${unsafeSVG(downIcon)}</span>
-            </div>
+            ${this.dropdownAnchor === 'button'
+              ? html`
+                  <slot
+                    name="button"
```
brian-patrick-3, line 305: Should we name this slot "anchor" maybe? I wonder if we could simplify it a bit, and just add a slot around the default anchor element.

src/components/reusable/dropdown/enhancedDropdown.stories.js
```diff
@@ -0,0 +1,630 @@
...
+  placeholder: '',
+  size: 'md',
+  inline: false,
+  name: 'example',
+  open: false,
+  required: false,
+  disabled: false,
+  hideTags: false,
+  hideLabel: false,
+  selectAll: false,
```
brian-patrick-3, line 66: This arg doesn't seem to be bound to the element prop.

src/components/reusable/dropdown/enhancedDropdown.stories.js
```diff
@@ -0,0 +1,630 @@
...
+import branchIcon from '@kyndryl-design-system/shidoka-icons/svg/monochrome/16/branch.svg';
+import downIcon from '@kyndryl-design-system/shidoka-icons/svg/monochrome/20/chevron-down.svg';
+
+export default {
+  title: 'Components/Dropdown/Enhanced Dropdown',
+  component: 'kyn-dropdown',
+  subcomponents: {
+    EnhancedDropdownOption: 'kyn-enhanced-dropdown-option',
+  },
+  argTypes: {
```
brian-patrick-3, line 21: Should include the ValidationArgs I added to the other form input elements just the other day.

src/components/reusable/dropdown/enhancedDropdown.stories.js
```diff
@@ -0,0 +1,630 @@
...
+        .value=${args.value}
+        @on-change=${(e) => {
+          const selectedValues = e.detail.value;
+          args.value = selectedValues;
+          action(e.type)({ ...e, detail: e.detail });
+        }}
+      >
+        <kyn-enhanced-dropdown-option value="1">
+          <span slot="title">Option 1</span>
+          <span slot="description"
```
brian-patrick-3, line 441: Do we need to include the description in the Tags that appear below the multi-select?

src/components/reusable/dropdown/enhancedDropdownOption.ts
```diff
@@ -0,0 +1,231 @@
...
+
+import checkIcon from '@kyndryl-design-system/shidoka-icons/svg/monochrome/16/check.svg';
+import clearIcon from '@kyndryl-design-system/shidoka-icons/svg/monochrome/16/close-simple.svg';
+
+import EnhancedDropdownOptionScss from './enhancedDropdownOption.scss?inline';
+
+/**
+ * Enhanced Dropdown option with rich content support.
+ * @fires on-click - Emits the option details to the parent dropdown. `detail:{ selected: boolean, value: string, origEvent: PointerEvent }`
+ * @fires on-remove-option - Emits the option that is removed.
```
brian-patrick-3, line 16: Probably should include the detail defs for this event too.

brian-patrick-3 (commented):
Additionally, there seems to be a z-index issue with the menu, when it opens upwards I can see the input label bleeding through.

Seems solid overall, had a couple suggestions to maybe simplify things. Not a huge fan of needing a separate option subcomponent, but understand why it might be needed.

src/components/reusable/dropdown/enhancedDropdown.stories.js
wbarbee: isn't that a lot of info for a tag?

brian-patrick-3: Right that was my thought, it is currently showing both title and description

wbarbee:
> Additionally, there seems to be a z-index or bg color issue with the menu, when it opens upwards I can see the input label bleeding through.
> 
> Seems solid overall, had a couple suggestions to maybe simplify things. Not a huge fan of having a separate option subcomponent, but understand why it might be needed.

Screenshot 2025-07-30 at 3 14 15 PM
Screenshot 2025-07-30 at 3 14 33 PM

i think the z-index issue is actually an opacity issue (30%) on the background-forms-default token

brian-patrick-3:
Shouldn't the enhanced option bg token be the same as the standard option bg token?

kumarivijeta:
Observations :

1. In light mode , No On hover effect on options.
2. The checkmark color shouldn’t be similar to standard dropdown ?  
3. In dark mode , On hover checkbox appearance differs from figma.
4. In Button Anchor Varient -  Button used AI styles , Can we have the default/secondary button style here, To maintain consistency ?
5. Can we remove repeated enhanced word from each varient name ?
Screenshot 2025-07-31 at 11 07 20 AM

wbarbee:
Thanks, Vijeta. I think that Diana designed this component primarily for use in the context of AI agentic tools, so my guess is that this is why the designs might depart a bit from the generic dropdown.

So 2 and 4 are examples of this.

I've asked Diana about the hover token and also to reconcile the designs with our current checkbox hover effect styles, since we don't want to modify those for this specific case.

But I removed 'Enhanced' from each story title, good point.

kumarivijeta (approved):

brian-patrick-3 (approved):

PR #552, opened 2025-08-04
fix(multi input field): better api response handling, enhancement

wbarbee:
## Summary

Request in the 'General' chat -- a scenario where more control would be required on the part of the consuming dev to mark individual tags based on a server response, so that some tags render in red (errors) and others in spruce (success).

> HI Team
> Good Morning
>  
> We are trying to integrate this "kyn-multi-input-field" shidoka component for multi input emails
> https://shidoka-applications.netlify.app/?path=/docs/components-multi-input-field--docs
>  
> We have used the inputType attribute as Email and used the pattern attribute for email pattern validation and its working fine to validation email in front end.
>  
> But problem is that let's say if we enter an email called "test@google.com" which is valid based on our pattern attrubute then we call one server side api and server return that email is already exist in db then we need to show this email as red color in  "kyn-multi-input-field" component.
> storybook - Storybook
>  
> Same for multiple email we can send to the server if they are valid with pattern and after getting api response some of the emails may be already exist in db and some of not exist then we need to show some with error red tag and some as success with "spruce" color .
>  
> So after getting the api response we need to show error (red) tag or valid tag based on the response
>  
> Please suggest some way to overcome with this problem

Added a small public property `itemStatusMap` to the existing component so `{ email: 'error'|'success' }` that can be passed in and have each tag render the correct color.

New story: https://deploy-preview-552--shidoka-applications.netlify.app/?path=/story/components-multi-input-field--server-validated-emails

## Checklist

✅ Consistent API - invalidText works identically to textInput

✅ Complete validation states - supports success, error, and default states

✅ Error display - error icon only shows when there's an actual message

✅ Proper form integration - internal validity correctly reflects all error sources

✅ Priority handling - invalidText takes precedence over other validation messages

✅ Backward compatibility - existing implementations continue to work unchanged

✅ Documentation - JSDoc and Storybook examples

### DIFF
--- a/src/components/reusable/multiInputField/multiInputField.stories.js
+++ b/src/components/reusable/multiInputField/multiInputField.stories.js
@@ -259,9 +259,14 @@ CustomInvalidText.args = {
 };
 
 export const ServerValidatedEmails = () => {
-  const emails = ['validated@example.com', 'error@example.com'];
+  const emails = [
+    'validated@example.com',
+    'default@example.com',
+    'error@example.com',
+  ];
   const statusMap = {
     'validated@example.com': 'success',
+    'default@example.com': 'default',
     'error@example.com': 'error',
   };
 
@@ -271,7 +276,7 @@ export const ServerValidatedEmails = () => {
       .inputType=${'email'}
       .itemStatusMap=${statusMap}
       label="Server-Validated Emails"
-      caption="Shows different validation states: success (spruce), default (neutral), and error (red)."
+      caption="Shows different validation states: success (spruce), default (default), and error (red)."
       placeholder="Emails from server"
       @on-change=${(e) => action('on-change')(e.detail)}
     >
--- a/src/components/reusable/multiInputField/multiInputField.ts
+++ b/src/components/reusable/multiInputField/multiInputField.ts
@@ -222,7 +222,7 @@ export class MultiInputField extends FormMixin(LitElement) {
     } else if (explicit === 'success') {
       return this._renderTagWithColor(item, index, 'spruce');
     } else if (explicit === 'default') {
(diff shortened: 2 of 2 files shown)


src/components/reusable/multiInputField/multiInputField.stories.js
```diff
@@ -240,3 +241,28 @@ WithCustomIcon.args = {
...
+    'hello@world.com': 'success',
+  };
+
+  return html`
+    <kyn-multi-input-field
+      .value=${emails.join(',')}
+      .inputType=${'email'}
+      .itemStatusMap=${statusMap}
+      label="Server-Validated Emails"
+      caption="Red
```
brian-patrick-3, line 259: `itemStatusMap` seems solid, but we may also want an `invalidText` prop or similar for custom error messages. Not sure caption works for that.

brian-patrick-3: Both of these things would probably also need to trigger the internal validity state of the input as well.

src/components/reusable/multiInputField/multiInputField.ts
```diff
@@ -92,6 +92,13 @@ export class MultiInputField extends FormMixin(LitElement) {
   @property({ type: String })
   accessor pattern: string | undefined = undefined;
 
+  /**
+   * Consumer-driven status map, e.g.
+   * { "foo@ex.com": "error", "bar@ex.com": "success" }
+   */
+  @property({ type: Object })
+  accessor itemStatusMap: Record<string, 'error' | 'success'> = {};
```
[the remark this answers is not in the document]
brian-patrick-3: Yeah I just realized "success" is basically our default state here

brian-patrick-3 (approved):

PR #555, opened 2025-08-05
fix(enhanced dropdown): chevron in ButtonAnchor story

wbarbee:
## Summary

Fix chevron in ButtonAnchor story

### DIFF
--- a/src/components/reusable/dropdown/dropdown.ts
+++ b/src/components/reusable/dropdown/dropdown.ts
@@ -1200,7 +1200,7 @@ export class Dropdown extends FormMixin(LitElement) {
     if (!root) return;
 
     if (changedProps.has('open')) {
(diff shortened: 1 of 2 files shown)


brian-patrick-3 (commented):
Hmm I find this almost kind of pointless to portray here, since the CSS would have to be manually added by the developer. Unless we can find a way to bake it in, it might just be better not to have this.

wbarbee:
> Hmm I find this almost kind of pointless to portray here, since the CSS would have to be manually added by the developer. Unless we can find a way to bake it in, it might just be better not to have this.

it was matching an original button example from Diana's original designs. really doesn't matter to me whether we have a chevron or not. i'll just remove it

brian-patrick-3:
I think it needs the chevron to indicate it's a dropdown, but adding external css to drive the rotate animation was problematic.

brian-patrick-3 (approved):

Issue #266, opened 2024-10-23
In the Number input, when the Label text is long, its showing in multiple lines.


srpriyankashetty:
**Before submitting a bug:**

- [yes] Is this reproducible in the latest version of Shidoka?
- [yes] Is this reproducible in the [Shidoka Storybook](https://kyndryl-design-system.github.io/shidoka-applications/)?
- [ ] Is this reproducible in a [Sandbox](https://codesandbox.io)?
- If it's not reproducible in any of the above scenarios, please do not open a bug report as it is an application-level issue.

**Describe the bug**

In the Number input, when the Label text is long, its showing in multiple lines.

**To Reproduce**

1. Go to Shidoka Storybook
2. Click on Number input
3. Scroll down to 'Label' - enter a long text
4. See the label is showing in multiple lines

**Expected behavior**

The text must show in one line

**Screenshots**

![image](https://github.com/user-attachments/assets/c851d0e2-0bc8-4c87-be45-e83d325a22a0)

**Environment**

- OS: [e.g. iOS] Windows
- Browser [e.g. chrome, safari] Chrome, Edge, Firefox

**Additional context**

Cross checked for Text area, but the text box is also expanding based on the label provided.

[AB#2096295](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/2096295)

brian-patrick-3:
@kyndryl-design-system/designers need to know if the number input should have a maximum width or something before proceeding.

wbarbee:
resolved in #562

PR #562, opened 2025-08-12
fix(number input): remove width constraints, fill parent containing div

wbarbee:
## Summary

min-/max-/width were set on numberInput.scss in such a way that it was not filling the width of the containing parent div like other form inputs.

1. Adjusted styles to remove constraints and set max-width in the two storybook instances where `<kyn-number-input>` is used.

2. Addressed issue with labels wrapping -- consuming dev can now control the width of the component itself (which dictates the width of the label) and the width of the internal input, separately.

## Github Issue

#266

## Checklist

### DIFF
--- a/src/components/reusable/numberInput/numberInput.stories.js
+++ b/src/components/reusable/numberInput/numberInput.stories.js
@@ -57,8 +57,12 @@ const args = {
   },
 };
 
-export const NumberInput = {
-  args,
+export const DefaultFullWidth = {
+  args: {
+    ...args,
+    label:
+      'Extra Extra Extra Extra Extra Extra Extra Extra Extra Extra Extra Extra Long Label',
+  },
   render: (args) => {
     return html`
       <kyn-number-input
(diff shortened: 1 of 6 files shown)


brian-patrick-3:
We should also address #266 here. As yet to be determined, we want kyn-number-input to have a mix-width of 100%, but internally the input itself might need a smaller max width.

src/components/reusable/numberInput/numberInput.stories.js
```diff
@@ -24,6 +24,16 @@ export default {
     max: {
       control: { type: 'number' },
     },
+    '--kyn-number-input-inner-max-width': {
+      control: { type: 'text' },
+      table: { category: 'CSS Custom Properties' },
+      defaultValue: '200px',
```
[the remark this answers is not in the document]
brian-patrick-3: Gotcha, it probably defaults to Object if it doesn't know the type. If the control was actually wired up that might also correct it, or alternatively making the control disabled or readonly if unused.

src/components/reusable/numberInput/numberInput.ts
```diff
@@ -21,10 +21,14 @@ const _defaultTextStrings = {
...
 /**
  * Number input.
+ *
  * @fires on-input - Captures the input event and emits the value and original event details.`detail:{ value: number }`
  * @slot tooltip - Slot for tooltip.
  * @attr {string} [name=''] - The name of the input, used for form submission.
  * @attr {string} [invalidText=''] - The custom validation message when the input is invalid.
+ *
+ * @cssprop --kyn-number-input-inner-max-width - Maximum width of the number input inner container.
+ * @cssprop --kyn-number-input-inner-min-width - Minimum width of the number input inner container.
```
brian-patrick-3, line 31: I think you could also specify the default here, ex: `@cssprop [--kyn-number-input-inner-min-width=0px]`

wbarbee: yeah, that works too. i'll correct this in my next PR. reads the same from the consuming dev's perspective, but cleaner on our end. good to know

PR #563, opened 2025-08-18
feat(accordion with actions): ability to add actions for accordion

shivarajkv:
## Summary

Ability to perform actions on accordion items, also provide user an option to place the toggle button either extreme left or right side of the title. 

Usecase:

In KIRS service. We have groups which contains a number of workflows. Users can perform operations at group level and also some operations at workflow level. At the same time we have a co-relation between these two entities. Some operations cannot be performed at group level when workflows are at certain state. So we need a component where user can view the state of the workflow and looking at that user can perform operations at group level. Also we needed a mechanism where user need not to load all the workflows eagerly. So accordion might be a correct fit for it. But in current accordion we cannot perform any actions. So an accordion where user is provided with the ability to perform certain operation at the accordion title level and also view the items under the title and if needed user can have interaction at the accordion details level will useful.

Sample image example:
image
 

## Notes

1. Provided argument in accordion component to place toggle button either to the left or right of the accordion title. Since option is provided at the parent element. It will be applicable to all the accordion items.
2. Added a slot for user to add overflow menu at the right side of the title, if user wishes to perform any action at the item level.

## Checklist

## Testing Instructions

Added a story 'Accordion With Actions'. 

## Screenshots

image


shivarajkv:
> Thanks for submitting this contribution! While we await time for the design team to review things, just a few initial observations on this PR, @shivarajkv.
> 
> Taking care of those items would be good in preparing for the team to begin formal review. Thanks!
> 
>     1. Could you please provide a use case justification in your summary above? Was there a concrete need for this utility in a product/project that you have been working on internally for Kyndryl? Please elaborate so that we can incorporate that into our own analysis. There are always going to be variants and functionality that can be added onto the existing components, however, in the interest of keeping the DS as lightweight as possible, we always want to make sure that there's a true need for what is being built.
> 
>     2. The general spacing of these right aligned elements -- whether it is overflow menu (...) only or the overflow + chevron icon -- should have some more breathing room from the far right edge of the accordion IMO.
> 
> 
> Screenshot 2025-08-21 at 2 11 17 PM
> 
> You can see what happens when the overflow is in focused state: Screenshot 2025-08-21 at 2 41 05 PM
> 
>     3. The chevrons (down arrows) should probably align vertically, regardless of whether an overflow is present in that accordion item (idk, this is a design preference that the design team can probably weigh in on). It's good that you're showing examples of both cases.
> 
> 
> Screenshot 2025-08-21 at 2 11 34 PM
> 
>     4. If this is simply toggling between two values (in this case "left" or "right"), better to set that to either:
> 
> 
> ```
> control: { type: 'select' },
> options: ['left', 'right']
> ```
> 
> or
> 
> ```
> control: { type: 'radio' }
> options: ['left', 'right']
> ```
> 
> Screenshot 2025-08-21 at 2 15 31 PM

image
image
@wbarbee Updated story book with options and padding issue also resolved.

brian-patrick-3 (commented):
There are some problems with this change:
1. The Accordion Item clickable area has changed. There are parts of the accordion item that are no longer clickable to expand it.
2. The accordion item focus state is broken compared to current version.
3. The overflow menu cannot be nested within the clickable area, or it creates a nested interactive accessibility violation. I expect this will appear after correcting item 1.
4. The arrow rotate transition is no longer centered because the `.expand-icon` container is no longer square.
5. There needs to be a gap between the arrow and the overflow menu.
6. I'm not sure we want to support the left togglePosition. Will need to check with design team.

PR #588, opened 2025-09-09
feat(overflowMenu): add expandable menu item

wbarbee:
## Summary

**Primary**: Add overflowMenuItem `nested` property to allow for a nested overflow menu. This is inferred by the presence of the nested overflow menu:

```
    <kyn-overflow-menu
      @on-click=${(e) => action('on-click')({ ...e, detail: e.detail })}
      ?open=${args.open}
      ?anchorRight=${args.anchorRight}
      ?verticalDots=${args.verticalDots}
      kind=${args.kind}
      ?fixed=${args.fixed}
      assistiveText=${args.assistiveText}
      .backButtonText=${args.backButtonText}
      width=${args.width}
      nestedWidth=${args.nestedWidth}
    >
      <kyn-overflow-menu-item>Option 1</kyn-overflow-menu-item>

      <kyn-overflow-menu-item>
        Nested Option
        <div slot="submenu">
          <kyn-overflow-menu-item
            @on-click=${(e) => action(e.type)({ ...e, detail: e.detail })}
            >Nested Option 1</kyn-overflow-menu-item
          >
          <kyn-overflow-menu-item
            @on-click=${(e) => action(e.type)({ ...e, detail: e.detail })}
            >Nested Option 2</kyn-overflow-menu-item
          >
          <kyn-overflow-menu-item
            destructive
            description="Destructive"
            @on-click=${(e) => action(e.type)({ ...e, detail: e.detail })}
            >Nested Option 3</kyn-overflow-menu-item
          >
        </div>
      </kyn-overflow-menu-item>
    </kyn-overflow-menu>
```

**Additional**:

1) refactor keypress logic for better parent and child (nested) overflow menu navigation and selection
2) reduce top and bottom padding to size overflow menu items to 'SM' dimensions (updated in Figma as well)
3) previously, hover state color token was not being applied to `...` in truncated overflow menu option. this has been fixed
4) add properties `linkWidths`, `nestedWidth`, `width` so that devs can fine-tune control over main (parent) overflow menu width and nested (child) overflow menu width, with `'match-parent'` as the default

`width`: sets the parent menu’s box size. It affects trigger alignment, anchoring (fixed/absolute left/right math), overflow clipping, and is the reference when you want children to mirror it.

`nestedWidth`: sets the submenu’s box size, independently. It can be:
- a fixed CSS length (`'280px'`, `'22rem'`, '%’) for bespoke child sizing,
- `'auto'`/`null` to let content define it,
- `'match-parent'` to mirror the parent’s computed px width (not just the token), preserving visual parity even if the parent width comes from CSS vars/responsive rules.

Having both cleanly separates parent layout from child layout, while still allowing the common “same width as parent” case via `'match-parent'`.

[AB#2477648](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/2477648)

## Figma Link

[Figma](https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library?node-id=7192-23459&p=f&m=dev)

## Checklist

#### Default Nested Overflow Menu
Screenshot 2025-09-09 at 3 07 18 PM

#### Nested Menu = Main Width
Screenshot 2025-09-09 at 3 07 27 PM

#### Nested Menu ≠ Main Width
Screenshot 2025-09-09 at 3 07 34 PM

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -552,496 +552,6 @@
     },
     {
       "kind": "javascript-module",
-      "path": "src/components/global/localNav/index.ts",
-      "declarations": [],
-      "exports": [
-        {
-          "kind": "js",
-          "name": "LocalNav",
-          "declaration": {
-            "name": "LocalNav",
-            "module": "./localNav"
-          }
-        },
-        {
-          "kind": "js",
-          "name": "LocalNavLink",
-          "declaration": {
-            "name": "LocalNavLink",
-            "module": "./localNavLink"
-          }
-        },
-        {
-          "kind": "js",
-          "name": "LocalNavDivider",
-          "declaration": {
-            "name": "LocalNavDivider",
-            "module": "./localNavDivider"
-          }
-        }
-      ]
-    },
-    {
-      "kind": "javascript-module",
-      "path": "src/components/global/localNav/localNav.ts",
-      "declarations": [
-        {
-          "kind": "class",
-          "description": "The global Side Navigation component.",
-          "name": "LocalNav",
-          "slots": [
-            {
-              "description": "The default slot, for local nav links.",
-              "name": "unnamed"
-            },
-            {
-              "description": "Slot for a search input",
-              "name": "search"
-            }
-          ],
-          "members": [
-            {
-              "kind": "field",
-              "name": "pinned",
-              "type": {
-                "text": "boolean"
-              },
-              "default": "false",
-              "description": "Local nav pinned state.",
-              "attribute": "pinned"
-            },
-            {
-              "kind": "field",
-              "name": "textStrings",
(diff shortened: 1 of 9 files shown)


wbarbee:
> Another thought about open on hover/focus. If we do that should it also close (with delay) on mouseleave/blur, and/or if another submenu gets opened?

>Need to think of a solution for mobile that prevents the submenu from opening offscreen. One potential idea would be to layer the submenu on top of the parent menu and add some way to go back. Could do this for large screens too and it would also solve item 4.

imo, closing on delay with mouseleave/blur gets a bit messy. maybe we just close on selection, off click or escape key press like we currently have on the main/parent overflow menu?

i ran by this opening to the side design by marco yesterday and it seems like it works and maybe looks better on larger screens, at least. i agree, having the submenu open on top of the parent menu on smaller screens is best.

PR #655, opened 2025-11-18
fix(dropdownOption): checkbox multi-select

will-barbee-kyndryl:
## Summary

Checkbox not responding to click selections, only clicking on dropdown option text.

`notFocusable`, when true, should not disabled the checkbox. this prevents it from accurately, visually reflecting the `selected` state of the checkbox

Screenshot 2025-11-18 at 3 59 01 PM

Screenshot 2025-11-18 at 4 06 31 PM

### DIFF
--- a/src/components/reusable/dropdown/dropdownOption.ts
+++ b/src/components/reusable/dropdown/dropdownOption.ts
@@ -129,6 +129,8 @@ export class DropdownOption extends LitElement {
                   ?readonly=${!this.disabled && this.readonly}
                   notFocusable
                   .indeterminate=${this.indeterminate}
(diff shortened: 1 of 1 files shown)


brian-patrick-kyndryl:
> i can locally but not in the deploy preview

Well that's odd, but shouldn't push it through without knowing why and if it will do the same when consumed.

brian-patrick-kyndryl:
> found the real culprit. now it should be working in chrome and firefox on the deploy preview

While the checkboxes are clickable now, it's not emitting the dropdown on-change event when clicking the checkboxes directly.

will-barbee-kyndryl:
i'll see you in the new year

Issue #706, opened 2026-01-16
DatePicker:  DatePicker With Time and DateRangePicker With Time


kumari-vijeta-kyndryl:
**Before submitting a bug:**

- [x] Is this reproducible in the latest version of Shidoka?
- [x] Is this reproducible in the [Shidoka Storybook](https://shidoka-applications.netlify.app)?
- [ ] Is this reproducible in a [Sandbox](https://codesandbox.io)?
- If it's not reproducible in any of the above scenarios, please do not open a bug report as it is an application-level issue.

**Describe the bug**

Number is not getting updated when using up/down arrows controls inside date/daterange timepicker box. applicable for daterangepicker as well.

**To Reproduce**

1. Go to [Date](https://shidoka-applications.netlify.app/?path=/story/components-date-picker--date-with-time)
2. Click on calendar icon
3. Change the time using arrow controls in number input.
4. Observe the behaviour after some clicks number controls not responding number update.

**Expected behavior**

Time should get updated and reflected same in date input.

**Screenshots**

<img width="728" height="641" alt="Image" src="https://github.com/user-attachments/assets/67d085d7-de26-4b9f-b7db-ae191011b107" />

https://github.com/user-attachments/assets/595b5f07-be6c-4913-a01e-d2343c1320f4

https://github.com/user-attachments/assets/911d918e-e644-4c49-bbe2-a39d38ffa3b1

**Environment**

- OS
- Browser  - chrome, safari
- Version -latest

[AB#2709612](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/2709612)

PR #707, opened 2026-01-16
fix(bug-706): date and date time picker issues

will-barbee-kyndryl:
### Issue 1: `TypeError: i.getTime is not a function`

In the `updated()` lifecycle method, the code was calling `.getTime()` on `currentDates[0]` and `currentDates[1]` without verifying they were actual `Date` objects. When flatpickr's time inputs trigger updates, it can sometimes pass values that aren't Date instances.

Fix Added proper type checking with `instanceof Date` before calling `.getTime()` (#706 ):

### Issue 2: Time Input Arrow Controls Not Updating Numbers
Number is not getting updated when using up/down arrow controls inside date/daterange timepicker box.

Cause: The error from Issue 1 was causing the update cycle to fail, making it appear that time values weren't updating when clicking the arrow controls.

Fix: Updated the `handleDateChange` method in both components to filter for valid Date objects before processing:

```
const validDates = selectedDates.filter(
  (d): d is Date => d instanceof Date && !isNaN(d.getTime())
);
```

## Checklist


kumari-vijeta-kyndryl:
> can you try again please, @kumari-vijeta-kyndryl? not able to reproduce it locally/on deploy preview in safari or chrome

Re-started chrome browser and tested. Fix resolves the issue.

kumari-vijeta-kyndryl (approved):

PR #833, opened 2026-04-23
fix(categorical-nav): more button truncation, link attribution issues

will-barbee-kyndryl:
## Summary

Fix the categorical nav `More` behavior for slotted categories so it consistently opens the clicked category and properly hides overflow links in the root view.

## Problem

In the slotted `kyn-header-categories` flow, clicking `More` could open the wrong category when `kyn-header-category` did not have an explicit `id`. The wrapper generated fallback category ids internally, but the child category emitted only the raw host `id`, causing the handler to fall back to the first category in some cases.

Separately, overflow links were marked with the `hidden` attribute in the root view, but `kyn-header-link` did not apply `display: none` for `[hidden]`, so truncated links could still render visibly.

## Solution

- Resolve and propagate a stable category id for slotted categories, even when no explicit `id` is provided.
- Emit that resolved id from the generated `More` action instead of relying only on the host attribute.
- Resolve the clicked slotted category from the event path before opening detail view.
- Make `kyn-header-link[hidden]` render as `display: none` so root-view truncation behaves visually as intended.
_**justificaition**_: visibility: hidden would hide the text, but it would still keep each truncated kyn-header-link in layout. In this menu that would cause the wrong behavior:
a. you’d get blank vertical gaps where the hidden links still take up space
b. the More link would not sit directly after the last visible link
column/masonry layout could still size itself around those “hidden” items
it would be hiding visually, not truly removing the overflow items from the root view

- Add a focused Storybook regression story covering truncated root view to full detail view for `More`.

## Validation

- Verified the regression story builds in Storybook.
- Ran the focused Storybook interaction test in Chromium for the `More` regression scenario.
- Confirmed no expected breaking API changes; this is a bug fix for existing behavior.

## Additional Issue (Resolved)

Reported by Vijeta during Internal Demo -- The local nav doesn’t currently reset transient state when the viewport crosses the 42rem breakpoint.

#834 

## Checklist

### DIFF
--- a/src/components/global/header/CategoricalNav.stories.js
+++ b/src/components/global/header/CategoricalNav.stories.js
@@ -1,6 +1,7 @@
 import { unsafeSVG } from 'lit-html/directives/unsafe-svg.js';
 import { html } from 'lit';
 import { useArgs } from 'storybook/preview-api';
+import { expect, userEvent, waitFor } from 'storybook/test';
 
 import './';
 import '../../reusable/button';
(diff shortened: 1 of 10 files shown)


src/components/global/header/CategoricalNav.stories.js
```diff
@@ -750,6 +751,105 @@ export const WithCategorizedNavManualHtml = {
...
+          serviceOperationsCategory.shadowRoot.querySelectorAll(
+            'kyn-header-link'
+          )
+        ).find((link) => link.textContent?.includes('More'))
+      ).toBeUndefined();
+    });
+  },
+};
+MoreClickUsesClickedCategoryRegression.storyName =
+  'Slotted More Regression: Truncated Root To Full Detail';
```
brian-patrick-kyndryl: Is this story specifically for running tests on? I'm wondering if we might want to mark it as such. Something so it's less confusing to devs looking at the docs?

brian-patrick-kyndryl (approved):
I just noticed in the Docs page for categorical navigation, the properties reference section is blank.

Code changes lgtm.

kumari-vijeta-kyndryl (approved):

src/components/global/header/CategoricalNav.stories.js
will-barbee-kyndryl: Yes, that’s a good call. This story is mainly a regression story for Karen to illustrate the interaction test rather than a normal docs/example story. I’ll mark it more clearly as test-only so it’s less confusing for people browsing the docs.

brian-patrick-kyndryl: We could even move it entirely to a Test section at the bottom of the sidebar. One of the reasons I've avoided adding tests to regular stories is because of how they automatically play out when a user views them.

will-barbee-kyndryl:
Screenshot 2026-04-23 at 8 50 50 AM

added a Tests section at the bottom of the sidebar populated with this story to start.

docs for Categorical Nav should be showing Properties, etc. now

thanks

PR #837, opened 2026-04-30
feat(workspaceSwitcher): convert from component to pattern

will-barbee-kyndryl:
Addresses [AB#2947353](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/2947353)

## Summary

- Convert `workspaceSwitcher` from a shipped custom-element implementation into a Storybook-first pattern under `src/stories/workspaceSwitcher/`.
- Remove the old runtime component surface:
  - delete `src/components/global/workspaceSwitcher/*` (except moving the example JSON fixture)
  - remove `WorkspaceSwitcher` exports from `src/index.ts`
  - regenerate `custom-elements.json` so the retired custom elements are no longer published
- Add new pattern artifacts:
  - `WorkspaceSwitcher.stories.js`
  - `Docs.mdx`
  - `WorkspaceSwitcher.test.stories.js`
  - `example_workspace_switcher_data.json`
- Preserve the existing visual/interaction model while moving control into app/story logic:
  - two-pane desktop layout
  - mobile root/detail drill-down
  - flyout mobile summary sync
  - account meta copy feedback
  - favorite toggles
  - launch indicator handling
- Tighten the pattern after audit feedback:
  - restore correct mobile account meta / `Workspaces` behavior in the header flyout example
  - fix launch/favorite icon placement and hover behavior
  - tighten spacing to better match the original component
  - make inactive mobile panes non-interactive
  - restore “opens in a new tab” assistive text on actionable rows
- Improve the example/story guidance:
  - `WithSearch` now actually filters right-pane account items
  - account-meta copy is modeled as a real action instead of fake link navigation
  - docs now frame flyout integration as Storybook-specific glue, not a stable public contract
  - replace the legacy `--kyn-workspace-switcher-max-height` control with a pattern-level `maxHeight` example arg


brian-patrick-kyndryl:
Should we treat this as a breaking change, and keep the old components in deprecated state?

will-barbee-kyndryl:
> Should we treat this as a breaking change, and keep the old components in deprecated state?

this console team is the only one implementing this component/pattern. technically, we should but i don't know that it's necessary

brian-patrick-kyndryl:
> this console team is the only one implementing this component/pattern. technically, we should but i don't know that it's necessary

Ok. My only other concern is the code sample for the actual workspace switcher is not visible in any of the code panels.

will-barbee-kyndryl:
> > this console team is the only one implementing this component/pattern. technically, we should but i don't know that it's necessary
> 
> Ok. My only other concern is the code sample for the actual workspace switcher is not visible in any of the code panels.

since there are a lot of styles baked into this pattern, I tried to make the Docs the source of truth and provide copyable snippets, but I'll try to make more available in the 'Code' tab

will-barbee-kyndryl:
> > this console team is the only one implementing this component/pattern. technically, we should but i don't know that it's necessary
> 
> Ok. My only other concern is the code sample for the actual workspace switcher is not visible in any of the code panels.
Screenshot 2026-04-30 at 8 11 23 AM

a lot of styles in the 'Code' tab now. might be a little overwhelming, but it is a somewhat complex pattern

brian-patrick-kyndryl:
Ah my comment was more about the HTML visibility, but the amount of styles needed to drive this do make it a bit problematic as a pattern

will-barbee-kyndryl:
> Ah my comment was more about the HTML visibility, but the amount of styles needed to drive this do make it a bit problematic as a pattern

well, either we go back to the drawing board and create a bunch of child components that assemble into this one specific thing or ... what?

brian-patrick-kyndryl:
Probably just have to make due with it like this for now, but I think the answer is ultimately that this is a Bridge pattern, not really a design system pattern. The template should live in Bridge somewhere.

brian-patrick-kyndryl (approved):

will-barbee-kyndryl:
> Probably just have to make due with it like this for now, but I think the answer is ultimately that this is a Bridge pattern, not really a design system pattern. The template should live in Bridge somewhere.

i've exposed more html in the Code tab. this shouldn't have been taken on for the design system in the first place, agreed

will-barbee-kyndryl:
maybe we have an 'Internal' Section down by the new 'Tests' section that will house our purely internal components and patterns

PR #856, opened 2026-05-19
fix(overflow): removed unused code

kumari-vijeta-kyndryl:
## Summary

Added selected state 

Link here (if applicable).

## Figma Link

[Figma](https://www.figma.com/design/qyPEUQckxj8LUgesi1OEES/Component-Library?node-id=7192-126161&m=dev)

## Checklist

## Screenshots

image

### DIFF
--- a/custom-elements.json
+++ b/custom-elements.json
@@ -427,677 +427,6 @@
     },
     {
       "kind": "javascript-module",
-      "path": "src/components/global/footer/footer.ts",
-      "declarations": [
-        {
-          "kind": "class",
-          "description": "The global Footer component.",
-          "name": "Footer",
-          "slots": [
-            {
-              "description": "Default slot, for links.",
-              "name": "unnamed"
-            },
-            {
-              "description": "Slot for the logo, will overwrite the default logo.",
-              "name": "logo"
-            },
-            {
-              "description": "Slot for the copyright text.",
-              "name": "copyright"
-            }
-          ],
-          "members": [
-            {
-              "kind": "field",
-              "name": "rootUrl",
-              "type": {
-                "text": "string"
-              },
-              "default": "'/'",
-              "description": "URL for the footer logo link. Should target the application home page.",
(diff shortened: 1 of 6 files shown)


brian-patrick-kyndryl:
What exactly is this for? Overflow menu is meant for actions/navigation, not selection, so I'm not really sure this is a valid use case.

diana-gutierrez-kyndryl (approved):

diana-gutierrez-kyndryl:
> What exactly is this for? Overflow menu is meant for actions/navigation, not selection, so I'm not really sure this is a valid use case.

The main use case is bulk selection across large datasets with pagination. I think originally from a UX perspective I was mostly reusing the menu states/anatomy from the overflow since it already solved a lot of the interaction behavior and we were not considering the selected state yet. But now that we’re talking about selected states for the selection scope, this may actually make more sense as a dropdown type pattern rather than a traditional overflow menu.

brian-patrick-kyndryl:
OH this is for the data table batch selection? Shouldn't that be a checkbox still? I don't think overflow menu works in that regard. It needs to show the checkbox selected and indeterminate states still.

One of their examples used for this feature request was Gmail. The problem is, even gmail's batch selection options only apply to the current page, not the entire dataset.

kumari-vijeta-kyndryl:
> OH this is for the data table batch selection? Shouldn't that be a checkbox still? I don't think overflow menu works in that regard. It needs to show the checkbox selected and indeterminate states still.
> 
> One of their examples used for this feature request was Gmail. The problem is, even gmail's batch selection options only apply to the current page, not the entire dataset.

Header Checkbox still be there along with overflow menu option for batch selection enabled. 

image



Other batch selection example where batch selection/deselection across page handles via prop https://ej2.syncfusion.com/react/demos/#/bootstrap5/grid/checkbox-selection .

kumari-vijeta-kyndryl:
Decided not to have selected state on overflow menu option. Hence, this PR is not required.

PR #862, opened 2026-05-28
docs: dashboard example pattern

will-barbee-kyndryl:
Partially addresses [AB#3016200](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/3016200)

## Summary

- Added a new `Patterns/Dashboard Example Pattern` Storybook story with a full Shidoka UI shell dashboard implementation.
- Reused existing Global Switcher and Workspace Switcher patterns, plus local nav, page title, KPI cards, widgets, charts, and footer.
- Extracted and instantiated the Kyndryl trellis SVG background treatment with responsive fade styling.
- Expanded the Storybook Code tab source to expose the dashboard structure, styles, nav, KPIs, and chart examples for consumers.
- Installed `@kyndryl-cto/shidoka-studio` via tarball file (for now) and used it to generate template dashboard code

## Checklist

## Testing Instructions

## Screenshots
Screenshot 2026-05-28 at 12 13 46 PM

Screenshot 2026-05-28 at 11 54 41 AM

Screenshot 2026-05-28 at 12 17 27 PM


marcolopez-kyndryl (approved):
Thank you!

package.json
```diff
@@ -55,6 +55,7 @@
     "@commitlint/cli": "^17.4.4",
     "@commitlint/config-conventional": "^17.4.4",
     "@custom-elements-manifest/analyzer": "^0.9.4",
+    "@kyndryl-cto/shidoka-studio": "file:kyndryl-cto-shidoka-studio-2.105.1.tgz",
```
brian-patrick-kyndryl: Assuming this is temporary until we have a proper location/installation method for shidoka studio?

kyndryl-cto-shidoka-studio-2.105.1.tgz
brian-patrick-kyndryl, line 1: Is this something we really want to commit to the repo's main branch, especially if it's temporary?

brian-patrick-kyndryl (commented):
I'm a bit out of the loop so may have some dumb questions, so please bear with me:

1. Assuming this is still proof of concept phase, why are we targeting the `main` branch?
2. I was under the impression Shidoka Studio was an MCP server, that it would be separate and standalone, why does it need to be installed as a dev dependency?
3. As far as I can tell, there is no added publishable feature here to mark this as a `feat`.
4. Does the dashboard example page freeze the browser and take a bit to render, or is it just me? Noticeable in Chrome, and more so in FF.
5. The local nav seems to be rendering on top of the header flyout.

src/stories/dashboardExamplePattern.stories.js
```diff
@@ -0,0 +1,1378 @@
...
+  </div>
+`;
+
+const renderChartWidget = (chart) => html`
+  <div
+    class="${chart.className} dashboard-chart-cell${chart.type === 'radar'
+      ? ' dashboard-chart-cell--radar'
+      : ''}"
+  >
+    <kyn-widget widgetTitle=${chart.title}>
```
brian-patrick-kyndryl: If using Widgets instead of Cards, shouldn't this use [gridstack](https://shidoka-applications.netlify.app/?path=/docs/components-layout-structure-widget-gridstack--docs) instead of a static grid?

src/stories/dashboardExamplePattern.stories.js
```diff
@@ -0,0 +1,1378 @@
...
+import '../components/global/footer';
+import '../components/reusable/badge';
+import '../components/reusable/button';
+import '../components/reusable/iconSelector';
+import '../components/reusable/link';
+import '../components/reusable/pagetitle';
+import '../components/reusable/tabs';
+import '../components/reusable/widget';
+import '@kyndryl-design-system/shidoka-charts/components/chart';
+import '@kyndryl-cto/shidoka-studio/styles/shell-overlays.css';
```
brian-patrick-kyndryl: What is this? I think the rogue header z-index is coming from here. Why would shidoka studio be serving css styles?

src/stories/dashboardExamplePattern.stories.js
```diff
@@ -0,0 +1,1378 @@
...
+  home: homeIcon,
+  launch: launchIcon,
+  recommend: starOutlineIcon,
+  'recommend-filled': starFilledIcon,
+  services: servicesIcon,
+  'user-settings': adminIcon,
+};
+
+const DASHBOARD_STYLES = /* css */ `
+  .ui-shell-dashboard-demo {
```
brian-patrick-kyndryl: I'm not sure how well all of this custom CSS will translate from this pattern to someone's actual implementation. Any ways we can reduce/simplify it?

src/stories/dashboardExamplePattern.stories.js
```diff
@@ -0,0 +1,1378 @@
...
+const stringifySourceProp = (value) => JSON.stringify(value, null, 2);
+
+const createChartSource = (chart) => `
+      <div class="${chart.className} dashboard-chart-cell${
+  chart.type === 'radar' ? ' dashboard-chart-cell--radar' : ''
+}">
+        <kyn-widget widgetTitle="${chart.title}">
+          <span slot="subtitle">${chart.subtitle}</span>
+          <div class="dashboard-chart-body">
+            <div class="dashboard-chart-host">
```
brian-patrick-kyndryl: [The widget is already designed to have charts dropped directly inside it](https://shidoka-applications.netlify.app/?path=/story/components-layout-structure-widget--with-chart). Why the extra containers/styles?

will-barbee-kyndryl:
> I'm a bit out of the loop so may have some dumb questions, so please bear with me:
> 
> 1. Assuming this is still proof of concept phase, why are we targeting the `main` branch?
> 2. I was under the impression Shidoka Studio was an MCP server, that it would be separate and standalone, why does it need to be installed as a dev dependency?
> 3. As far as I can tell, there is no added publishable feature here to mark this as a `feat`.
> 4. Does the dashboard example page freeze the browser and take a bit to render, or is it just me? Noticeable in Chrome, and more so in FF.
> 5. The local nav seems to be rendering on top of the header flyout.

Good questions.

1. Shidoka Studio is still in POC phase. I opened it against `main` because that’s the normal flow for this repo, but I’m not attached to that.

2. Shidoka Studio is just an MCP server. My understanding is that including it as a temporary tarball and referencing it as a dev dependency would only affect people working in this repo, not consumers of the published design system package. I added it thinking you and Vijeta might want to try it while we start building templates and larger patterns, but I’ve removed it from this PR.

3. I changed it from `feat` to `prototype`.

4. That’s not expected. I wasn’t seeing it locally in Chrome, but I reduced the chart load and turned off chart animation to make the story lighter. I’ll re-check Firefox too.

5. I fixed the stacking so the header flyouts render above the local nav.

believe i addressed everything else as well but LET ME KNOW

brian-patrick-kyndryl:
>     2. Shidoka Studio is just an MCP server. My understanding is that including it as a temporary tarball and referencing it as a dev dependency would only affect people working in this repo, not consumers of the published design system package. I added it thinking you and Vijeta might want to try it while we start building templates and larger patterns, but I’ve removed it from this PR.

Yeah ok, I guess I'm just wondering how we're planning for people to install and use shidoka studio, ex completely independently vs installing it as a dependency in an application. It seems to operate independently, so that's why I was unsure about the dependency method.

Ultimately for this PR then, all we're really pushing is a `docs` update containing this dashboard pattern?

src/stories/dashboardExamplePattern.stories.js
```diff
@@ -0,0 +1,1525 @@
...
+              <div
+                gs-id="${chart.id}"
+                class="grid-stack-item${
+                  chart.type === 'radar' ? ' dashboard-chart-item--radar' : ''
+                }"
+              >
+                <div class="grid-stack-item-content">
+                  <kyn-widget removeHeader>
+${indentSource(createChartHeaderSource(chart), 20)}
+                    <div class="dashboard-chart-plot">
```
brian-patrick-kyndryl: If the kd-chart is a direct child of kyn-widget, it's supposed to automatically detect it to hide the widget header in order to use the chart header in place.

will-barbee-kyndryl:
Still polishing the dashboard

src/stories/dashboardExamplePattern.stories.js
```diff
@@ -0,0 +1,1525 @@
...
+
+const renderChartHeader = (chart) => html`
+  <div class="dashboard-chart-header">
+    <div class="dashboard-chart-heading">
+      <div class="dashboard-chart-title kd-type--ui-02">${chart.title}</div>
+      <div class="dashboard-chart-description kd-type--ui-03">
+        ${chart.subtitle}
+      </div>
+    </div>
+    <kyn-widget-drag-handle
```
brian-patrick-kyndryl: Drag handle seems to not be slotting correctly.

will-barbee-kyndryl: still polishing

will-barbee-kyndryl: you can pause for a sec

will-barbee-kyndryl:
> > ```
> > 2. Shidoka Studio is just an MCP server. My understanding is that including it as a temporary tarball and referencing it as a dev dependency would only affect people working in this repo, not consumers of the published design system package. I added it thinking you and Vijeta might want to try it while we start building templates and larger patterns, but I’ve removed it from this PR.
> > ```
> 
> Yeah ok, I guess I'm just wondering how we're planning for people to install and use shidoka studio, ex completely independently vs installing it as a dependency in an application. It seems to operate independently, so that's why I was unsure about the dependency method.
> 
> Ultimately for this PR then, all we're really pushing is a `docs` update containing this dashboard pattern?

That part is still TBD and up to us. My current thinking is that a Node package is probably the best distribution model while the MCP is actively evolving, because it gives us versioning, reproducibility, and a clear install path for consuming repos.

I agree it is not a runtime dependency. Once development is complete, the MCP/package is not consumed by the application. But I think that is roughly analogous to any dev-time tooling like unit test libraries, inters, or codegen tools.  they help authors build and validate the app, but they're not shipped as part of the production runtime.

For this PR, yes, I think the scope is just the docs update containing the dashboard pattern, now that we've removed the MCP from the repo.

brian-patrick-kyndryl:
OK I think I understand the benefits of installing it as a dev dependency now. Yeah it wouldn't hurt anything to include the tarball temporarily, or we can just manually share it between us until it's published somewhere. I'm wondering if we'd want to also include any npm scripts in the package.json to execute it.

will-barbee-kyndryl:
> OK I think I understand the benefits of installing it as a dev dependency now. Yeah it wouldn't hurt anything to include the tarball temporarily, or we can just manually share it between us until it's published somewhere. I'm wondering if we'd want to also include any npm scripts in the package.json to execute it.

yes i have some `npx` install scripts that are included in the package itself. in the boilerplate repos i've created, i've set up easier local scripts. still working the kinks out in the install process while also improving the thing overall

will-barbee-kyndryl:
ok now the dashboard should be in better shape

brian-patrick-kyndryl (approved):

brian-patrick-kyndryl:
>     4. That’s not expected. I wasn’t seeing it locally in Chrome, but I reduced the chart load and turned off chart animation to make the story lighter. I’ll re-check Firefox too.

It may just be storybook itself choking on large/complex stories.

PR #878, opened 2026-06-11
fix: dashboard example widget alignment

will-barbee-kyndryl:
## Summary

Repair slight dashboard widget misalignment by standardizing gap between widgets

#### Before
image

#### After
Screenshot 2026-06-15 at 7 04 45 AM

### DIFF
--- a/src/stories/dashboardExamplePattern.stories.js
+++ b/src/stories/dashboardExamplePattern.stories.js
@@ -124,20 +124,20 @@ const DASHBOARD_PATTERN_STYLES = /* css */ `
     width: 100%;
     max-width: none;
     margin-inline: 0;
-    gap: 2rem;
+    gap: 32px;
   }
 
   .dashboard-kpi {
     min-height: 156px;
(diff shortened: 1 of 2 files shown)


src/stories/dashboardExamplePattern.stories.js
```diff
@@ -122,20 +122,22 @@ const DASHBOARD_PATTERN_STYLES = /* css */ `
     flex-wrap: wrap;
     align-items: stretch;
     width: 100%;
-    gap: 1.25rem;
+    max-width: none;
+    margin-inline: 0;
+    gap: 32px;
```
brian-patrick-kyndryl: Why are we overriding the out of the box grid utility classes in this section? kd-grid utility classes already specify the gap and everything else that should be needed, and would simplify the pattern removing custom css.

will-barbee-kyndryl: fixed

brian-patrick-kyndryl (approved):

Issue #879, opened 2026-06-15
split button: menu exceeds viewport bounds on right side


brian-patrick-kyndryl:
Hi team,

We’re seeing an issue with the **Shidoka split button** when it is placed near the **end (right edge) of the page width**.

**Problem:**

* When the split button dropdown is clicked in this position,  
   * it **adds a horizontal scroll to the page**, and  
   * the **dropdown content is not constrained within the viewport** (overflows outside the visible area).

**Expected Behavior:**

* The dropdown should **auto-adjust its position** (e.g., open towards left or reposition) to stay fully visible within the viewport.
* It should **not introduce any page-level scrolling**. 
 
[Microsoft Teams Message](https://teams.microsoft.com/l/message/19:m15yXA3kC7fl9TvnIML-6DxhfDtwyoLJMu0nVAeddXQ1@thread.tacv2/1781501580935)

[AB#3030643](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/3030643)

PR #880, opened 2026-06-15
fix(split-button): keep menu within viewport

will-barbee-kyndryl:
## Summary

Fixes two issues with the `kyn-split-btn` dropdown menu (bug-879):

1. **Viewport overflow / horizontal scroll** — When the split button was placed
   near the right edge of the page, opening the menu pushed content past the
   viewport and introduced horizontal page scroll. The menu now auto-repositions
   to stay fully visible.
2. **Invisible hover highlight** — Menu options had no visible hover/highlight
   state. They now match the `kyn-dropdown` menu exactly.

The split button uses its own custom-built menu (it does **not** consume
`kyn-dropdown`), so both fixes are contained within the split button component.

## Changes

**Viewport positioning (`splitButton.ts`, `splitButton.scss`)**
- Added a horizontal flip that mirrors the existing vertical (`_openUpwards`)
  logic. New `_openLeftwards` state anchors the menu to the button's right edge
  (`right: 0; left: auto`) when a left-aligned menu would overflow.
- Flip decision uses alignment-independent measurements
  (`wrapper.left + menuWidth` vs `document.documentElement.clientWidth`), so it's
  stable across opens and never triggers page scroll (clientWidth excludes the
  scrollbar).
- Added `max-width: calc(100vw - 16px)` (overridable via
  `--kyn-split-btn-menu-max-width`) so the menu can never exceed the viewport.

**Hover/highlight (`splitButtonOption.ts`, `splitButtonOption.scss`)**
- Root cause: `--kd-color-background-accent-subtle` resolves to `interactive-5`
  (#f6f7f9) — identical to the menu background `--kd-color-background-container-default`
  in light mode — so the hover was applied but invisible.
- Options now reuse the shared `common/scss/menu-item.scss` via the
  `.menu-item` / `.menu-item-inner-el` classes (same approach as
  `kyn-dropdown-option`), guaranteeing identical hover/highlight/selected/disabled
  states that stay in sync with the design system.

**Story**
- Added a "Near Viewport Edge" story under
  `Components/Buttons & Actions/Split Button` (alongside `Default` / `With Icon`)
  — a far-right button on a scrollable page that reproduces the original bug and
  demonstrates the fix. Tagged `!autodocs` so it stays in the sidebar without
  rendering its full-height scroll demo inline on the component's Docs page.

#879 

## Checklist

## Screenshots

Screenshot 2026-06-15 at 6 52 52 AM


brian-patrick-kyndryl (approved):

kumari-vijeta-kyndryl:
The fix looks good. But I am skeptical about having separate story.

kumari-vijeta-kyndryl (approved):

will-barbee-kyndryl:
> The fix looks good. But I am skeptical about having separate story.

migrated it to be nested under Split Button instead

PR #919, opened 2026-07-23
fix(side-drawer): support subtitle link

will-barbee-kyndryl:
## Summary
- Adds a `label` slot to `kyn-side-drawer` so the subtitle area supports rich content, including inline `kyn-link` components.
- Keeps `labelText` as a fallback for plain-string labels — existing usage is unchanged and non-breaking.
- Adds baseline alignment styling for slotted links and a new **Label Slot With Link** Storybook story demonstrating the pattern.
- Updates `custom-elements.json` to document the new slot.

## Test plan
- [x] Open Storybook → Side Drawer → **Label Slot With Link** and confirm the subtitle renders with an inline “Learn more” link
- [x] Verify existing drawers using only `labelText` still render the same plain-text subtitle
- [x] Confirm the drawer header hides the label region when both `labelText` and the `label` slot are empty
- [x] Click the link and confirm navigation / `on-click` fires as expected

[AB#3077856](https://dev.azure.com/Kyndryl/f7f7ab25-06ec-43dc-902d-dee2e157cebd/_workitems/edit/3077856)

### DIFF
--- a/src/components/reusable/sideDrawer/sideDrawer.ts
+++ b/src/components/reusable/sideDrawer/sideDrawer.ts
@@ -1,6 +1,12 @@
 import { unsafeSVG } from 'lit-html/directives/unsafe-svg.js';
 import { LitElement, html, css, unsafeCSS } from 'lit';
-import { customElement, property, query, state } from 'lit/decorators.js';
+import {
+  customElement,
+  property,
+  query,
+  queryAssignedElements,
+  state,
+} from 'lit/decorators.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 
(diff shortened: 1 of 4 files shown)


izahir-cascante-sandi-kyndryl (approved):
LGTM

src/components/reusable/sideDrawer/sideDrawer.ts
```diff
@@ -150,6 +165,14 @@ export class SideDrawer extends LitElement {
   @query('dialog')
   accessor _dialog!: any;
 
+  private get _showLabel() {
```
kumari-vijeta-kyndryl, line 169: Mark it as @internal/@ignore .

kumari-vijeta-kyndryl (approved):

